PR #6, opened 2019-01-30
Fix libraries in filter.go to reflect how GCE API list filtering works currently

rramkumar1:
Biggest changes:

1. Regex is represented by '~' rather than 'eq' 
2. int or bool equality is represented by "=" rather than "=="

Will only submit once I verify this works against GCE. 

/assign @bowei

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/k8s-cloud-provider/6)

### DIFF
--- a/pkg/cloud/filter/filter.go
+++ b/pkg/cloud/filter/filter.go
@@ -43,17 +43,17 @@ var (
 	None *F
 )
 
-// Regexp returns a filter for fieldName matches regexp v.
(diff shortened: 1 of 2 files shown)


rramkumar1:
/close

Apparently, regex is only supported via the gcloud interface so this PR would not work for any other clients.

rramkumar1:
Reopening since these changes still help in supporting regex for local matching.

Issue #11, opened 2019-05-06
TestFilterToString is Failing


spencerhance:
Test fails when running `make` on master

**Output:**
```
--- FAIL: TestFilterToString (0.00s)
    filter_test.go:42: filter &filter.F{predicates:[]filter.filterPredicate{filter.filterPredicate{fieldName:"field1", op:0, s:(*string)(0xc00009a500), i:(*int)(nil), b:(*bool)(nil)}, filter.filterPredicate{fieldName:"field2", op:0, s:(*string)(0xc00009a4f0), i:(*int)(nil), b:(*bool)(nil)}}} String() = "\"(field1 ~ abc) (field2 ~ def)\"", want "(field1 ~ abc) (field2 ~ def)"
    filter_test.go:42: filter &filter.F{predicates:[]filter.filterPredicate{filter.filterPredicate{fieldName:"field1", op:0, s:(*string)(0xc00009a510), i:(*int)(nil), b:(*bool)(nil)}, filter.filterPredicate{fieldName:"field2", op:3, s:(*string)(nil), i:(*int)(0xc0000a60f8), b:(*bool)(nil)}}} String() = "\"(field1 ~ abc) (field2 != 17)\"", want "(field1 ~ abc) (field2 != 17)"
    filter_test.go:42: filter &filter.F{predicates:[]filter.filterPredicate{filter.filterPredicate{fieldName:"field1", op:0, s:(*string)(0xc00009a520), i:(*int)(nil), b:(*bool)(nil)}, filter.filterPredicate{fieldName:"field2", op:2, s:(*string)(nil), i:(*int)(0xc0000a6100), b:(*bool)(nil)}}} String() = "\"(field1 ~ abc) (field2 = 17)\"", want "(field1 ~ abc) (field2 = 17)"
E0506 16:53:36.108929   23308 filter.go:245] Match regexp "a(((" is invalid: error parsing regexp: missing closing ): `a(((`
FAIL
```

spencerhance:
Introduced by https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/6

PR #12, opened 2019-05-21
Revert String() change to escaped string

spencerhance:
Revert one line change from https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/6 so that the tests pass again.

Fixes https://github.com/GoogleCloudPlatform/k8s-cloud-provider/issues/11


spencerhance:
cc: @rramkumar1

PR #16, opened 2019-06-21
Add some more hooks

spencerhance:
Includes hooks for ILB and for https://github.com/kubernetes/ingress-gce/pull/781

### DIFF
--- a/pkg/cloud/mock/mock.go
+++ b/pkg/cloud/mock/mock.go
@@ -457,6 +457,26 @@ func UpdateRegionBackendServiceHook(ctx context.Context, key *meta.Key, obj *ga.
 	return nil
 }
 
+// UpdateRegionBackendServiceHook defines the hook for updating a Region
+// BackendsService. It replaces the object with the same key in the mock with
+// the updated object.
+func UpdateAlphaRegionBackendServiceHook(ctx context.Context, key *meta.Key, obj *ga.BackendService, m *cloud.MockAlphaRegionBackendServices) error {
+	_, err := m.Get(ctx, key)
+	if err != nil {
+		return &googleapi.Error{
+			Code:    http.StatusNotFound,
+			Message: fmt.Sprintf("Key: %s was not found in RegionBackendServices", key.String()),
+		}
+	}
+
+	obj.Name = key.Name
+	projectID := m.ProjectRouter.ProjectID(ctx, "ga", "backendServices")
+	obj.SelfLink = cloud.SelfLink(meta.VersionAlpha, projectID, "backendServices", key)
+
+	m.Objects[*key] = &cloud.MockRegionBackendServicesObj{Obj: obj}
+	return nil
+}
+
 // UpdateBackendServiceHook defines the hook for updating a BackendService.
 // It replaces the object with the same key in the mock with the updated object.
 func UpdateBackendServiceHook(ctx context.Context, key *meta.Key, obj *ga.BackendService, m *cloud.MockBackendServices) error {
@@ -519,10 +539,7 @@ func UpdateBetaBackendServiceHook(ctx context.Context, key *meta.Key, obj *beta.
 func UpdateURLMapHook(ctx context.Context, key *meta.Key, obj *ga.UrlMap, m *cloud.MockUrlMaps) error {
 	_, err := m.Get(ctx, key)
 	if err != nil {
-		return &googleapi.Error{
-			Code:    http.StatusNotFound,
-			Message: fmt.Sprintf("Key: %s was not found in UrlMaps", key.String()),
-		}
+		return err
 	}
 
 	obj.Name = key.Name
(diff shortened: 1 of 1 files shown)


pkg/cloud/mock/mock.go
```diff
@@ -457,6 +457,26 @@ func UpdateRegionBackendServiceHook(ctx context.Context, key *meta.Key, obj *ga.
 	return nil
 }
 
+// UpdateRegionBackendServiceHook defines the hook for updating a Region
+// BackendsService. It replaces the object with the same key in the mock with
+// the updated object.
+func UpdateAlphaRegionBackendServiceHook(ctx context.Context, key *meta.Key, obj *ga.BackendService, m *cloud.MockAlphaRegionBackendServices) error {
+	_, err := m.Get(ctx, key)
+	if err != nil {
```
bowei: Through the error unless it is actually a NotFound errors

pkg/cloud/mock/mock.go
```diff
@@ -552,6 +572,25 @@ func UpdateAlphaURLMapHook(ctx context.Context, key *meta.Key, obj *alpha.UrlMap
 	return nil
 }
 
+// UpdateAlphaURLMapHook defines the hook for updating an alpha UrlMap.
+// It replaces the object with the same key in the mock with the updated object.
+func UpdateAlphaRegionURLMapHook(ctx context.Context, key *meta.Key, obj *alpha.UrlMap, m *cloud.MockAlphaRegionUrlMaps) error {
+	_, err := m.Get(ctx, key)
+	if err != nil {
```
bowei: see comment above

pkg/cloud/mock/mock.go
```diff
@@ -566,6 +605,34 @@ func SetTargetGlobalForwardingRuleHook(ctx context.Context, key *meta.Key, obj *
 	return nil
 }
 
+// SetTargetForwardingRuleHook defines the hook for setting the target proxy for a ForwardingRule.
+func SetTargetForwardingRuleHook(ctx context.Context, key *meta.Key, obj *ga.TargetReference, m *cloud.MockForwardingRules) error {
+	fw, err := m.Get(ctx, key)
+	if err != nil {
+		return &googleapi.Error{
```
bowei: see above

pkg/cloud/mock/mock.go
```diff
@@ -594,6 +661,34 @@ func SetURLMapTargetHTTPProxyHook(ctx context.Context, key *meta.Key, ref *ga.Ur
 	return nil
 }
 
+// SetURLMapTargetHTTPProxyHook defines the hook for setting the url map for a TargetHttpProxy.
+func SetURLMapTargetHTTPSProxyHook(ctx context.Context, key *meta.Key, ref *ga.UrlMapReference, m *cloud.MockTargetHttpsProxies) error {
+	tp, err := m.Get(ctx, key)
+	if err != nil {
```
bowei: see above

pkg/cloud/mock/mock.go
```diff
@@ -594,6 +661,34 @@ func SetURLMapTargetHTTPProxyHook(ctx context.Context, key *meta.Key, ref *ga.Ur
...
+	}
+
+	tp.UrlMap = ref.UrlMap
+	return nil
+}
+
+// SetURLMapTargetHTTPProxyHook defines the hook for setting the url map for a TargetHttpProxy.
+func SetURLMapAlphaRegionTargetHTTPSProxyHook(ctx context.Context, key *meta.Key, ref *alpha.UrlMapReference, m *cloud.MockAlphaRegionTargetHttpsProxies) error {
+	tp, err := m.Get(ctx, key)
+	if err != nil {
```
bowei: see above

pkg/cloud/mock/mock.go
spencerhance: Okay I updated the error handling, not sure if that's what you had in mind

pkg/cloud/mock/mock.go
spencerhance: done

pkg/cloud/mock/mock.go
spencerhance: done

pkg/cloud/mock/mock.go
spencerhance: done

pkg/cloud/mock/mock.go
spencerhance: done

bowei:
probably just pass err unchanged back to caller? It's up to the mock (and users of the mock) to use the right type of error

spencerhance:
@bowei Makes sense to me - fixed in latest push.

pkg/cloud/mock/mock.go
bowei: you need to fix this as well?

spencerhance: Good catch.  Since i'm making this change anyway I can also just update the other errors.

spencerhance: Done

PR #22, opened 2019-08-01
Add UpdateAlphaRegionHealthCheckHook() mock

spencerhance:


pkg/cloud/mock/mock.go
```diff
@@ -447,6 +447,22 @@ func UpdateAlphaHealthCheckHook(ctx context.Context, key *meta.Key, obj *alpha.H
 	return nil
 }
 
+// UpdateAlphaRegionHealthCheckHook defines the hook for updating an alpha HealthCheck.
+// It replaces the object with the same key in the mock with the updated object.
+func UpdateAlphaRegionHealthCheckHook(ctx context.Context, key *meta.Key, obj *alpha.HealthCheck, m *cloud.MockAlphaRegionHealthChecks) error {
+	_, err := m.Get(ctx, key)
```
rramkumar1: I think you can do:

if _, err := m.Get(ctx, key); err != nil {
  return err
}

But, it's not really consistent in this file so up to you

spencerhance: Okay I switched it

rramkumar1:
LGTM

Assigning @bowei for final approval.

PR #23, opened 2019-08-12
fix Mock for AggregatedList

freehan:
The current mock for AggregatedList mock return the zone/region name as key directly. The real GCE API return key in following format `zones/{zone}`, `regions/{region}` or `global`

### DIFF
--- a/pkg/cloud/gen.go
+++ b/pkg/cloud/gen.go
@@ -2385,7 +2385,6 @@ func (m *MockAddresses) AggregatedList(ctx context.Context, fl *filter.F) (map[s
 	objs := map[string][]*ga.Address{}
 	for _, obj := range m.Objects {
 		res, err := ParseResourceURL(obj.ToGA().SelfLink)
-		location := res.Key.Region
 		if err != nil {
 			klog.V(5).Infof("MockAddresses.AggregatedList(%v, %v) = nil, %v", ctx, fl, err)
 			return nil, err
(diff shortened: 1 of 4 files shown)


pkg/cloud/gen.go
```diff
@@ -36,6 +36,15 @@ import (
 	ga "google.golang.org/api/compute/v1"
 )
 
+const (
+	// regionalKeyFmt is the key format of regional resources in the AggregateList response
+	regionalKeyFmt = "regions/%v"
+	// regionalKeyFmt is the key format of zonal resources in the AggregateList response
+	zonalKeyFmt = "zones/%v"
```
bowei: %s

pkg/cloud/gen.go
```diff
@@ -36,6 +36,15 @@ import (
 	ga "google.golang.org/api/compute/v1"
 )
 
+const (
+	// regionalKeyFmt is the key format of regional resources in the AggregateList response
+	regionalKeyFmt = "regions/%v"
```
bowei: %s

bowei: instead of separating out the fmt string, you should probably just do:
```
func aggregatedListKey(k meta.Key) string {
  switch ...
    case ...: return sprintf...
}
```

freehan:
Fixed

freehan:
Moved the aggregatedListKey to util and added unit test

PR #24, opened 2019-08-19
Fix libraries in filter.go to use the correct operators supported by GCE

prameshj:
Reverts GoogleCloudPlatform/k8s-cloud-provider#6

Most GCE APIs don't support regexp filter, hence reverting this.

Picked this up in k/k and Routes Controller started failing when listing routes with:

GCERoutes.List(context.Background.WithDeadline(2019-08-19 23:56:59.880823189 +0000 UTC m=+5923.032348891 [59m59.8650081s]), ..., (name ~ e2e-test-.*) (network ~ https://www.googleapis.com/compute/v1/projects/pavithrar-k8s-dev/global/networks/e2e-test) (description ~ k8s-node-route)) = <nil>, googleapi: Error 400: Invalid value for field 'filter': '(name ~ e2e-test-pavithrar-.*) (network ~ https://www.googleapis.com/compute/v1/projects/pavithrar-k8s-dev/global/networks/e2e-test-pavithrar) (description ~ k8s-node-route)'. Invalid list filter expression., invalid

The GCE APIs only support filter of the form "name = value" and not regexp. gcloud seems to be implementing regex filter and invoking API to get the full list.
ingress-gce uses composite types for most resources and no filter so far, which is probably why it has not broken any ingress tests.

### DIFF
--- a/pkg/cloud/filter/filter.go
+++ b/pkg/cloud/filter/filter.go
@@ -43,32 +43,32 @@ var (
 	None *F
 )
 
-// Regexp returns a filter for fieldName ~ regexp v.
+// Regexp returns a filter for fieldName eq regexp v.
 func Regexp(fieldName, v string) *F {
 	return (&F{}).AndRegexp(fieldName, v)
 }
 
-// NotRegexp returns a filter for fieldName !~ regexp v.
+// NotRegexp returns a filter for fieldName ne regexp v.
 func NotRegexp(fieldName, v string) *F {
 	return (&F{}).AndNotRegexp(fieldName, v)
 }
 
-// EqualInt returns a filter for fieldName = v.
+// EqualInt returns a filter for fieldName eq v.
 func EqualInt(fieldName string, v int) *F {
 	return (&F{}).AndEqualInt(fieldName, v)
 }
 
-// NotEqualInt returns a filter for fieldName != v.
+// NotEqualInt returns a filter for fieldName ne v.
 func NotEqualInt(fieldName string, v int) *F {
 	return (&F{}).AndNotEqualInt(fieldName, v)
 }
 
-// EqualBool returns a filter for fieldName = v.
+// EqualBool returns a filter for fieldName eq v.
 func EqualBool(fieldName string, v bool) *F {
 	return (&F{}).AndEqualBool(fieldName, v)
 }
 
-// NotEqualBool returns a filter for fieldName != v.
+// NotEqualBool returns a filter for fieldName ne v.
 func NotEqualBool(fieldName string, v bool) *F {
 	return (&F{}).AndNotEqualBool(fieldName, v)
 }
@@ -199,13 +199,16 @@ func (fp *filterPredicate) String() string {
 	var op string
 	switch fp.op {
 	case regexpEquals:
-		op = "~"
+		// GCE API maps regexp comparison to 'eq'
+		op = "eq"
 	case regexpNotEquals:
-		op = "!~"
+		op = "ne"
+	// Since GCE API does not allow using a mix of 'eq' and '=' operators,
+	// we use 'eq' everywhere
 	case equals:
-		op = "="
+		op = "eq"
 	case notEquals:
-		op = "!="
+		op = "ne"
 	default:
 		op = "invalidOp"
 	}
--- a/pkg/cloud/filter/filter_resource_test.go
+++ b/pkg/cloud/filter/filter_resource_test.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2019 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package filter
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"golang.org/x/oauth2/google"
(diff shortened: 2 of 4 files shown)


prameshj:
/assign @bowei @rramkumar1 
@spencerhance

prameshj:
/hold

Actually we need a part of the changes in this PR. Equals needs to translate to "=" and NotEquals needs to translate to " != ".
 From the GCE API docs:

"A filter expression that filters resources listed in the response. The expression must specify the field name, a comparison operator, and the value that you want to use for filtering. The value must be a string, a number, or a boolean. The comparison operator must be either =, !=, >, or <."

This is the same for Instances, Routes, Addresses, ForwardingRules.. 
We can keep RegexpEquals as an option still, but not the default. I will work on the changes and submit for review.

prameshj:
> /hold
> 
> Actually we need a part of the changes in this PR. Equals needs to translate to "=" and NotEquals needs to translate to " != ".
> From the GCE API docs:
> 
> "A filter expression that filters resources listed in the response. The expression must specify the field name, a comparison operator, and the value that you want to use for filtering. The value must be a string, a number, or a boolean. The comparison operator must be either =, !=, >, or <."
> 
> This is the same for Instances, Routes, Addresses, ForwardingRules..
> We can keep RegexpEquals as an option still, but not the default. I will work on the changes and submit for review.

UPDATE:

Turns out the GCE documentation is a bit confusing here, I tried this out using the "Try this API" section here: https://cloud.google.com/compute/docs/reference/rest/v1/routes/list

1) Using "eq" for string equality/regex works, The following filter returned the correct matches:
'(name eq e2e-test-pavithrar-.*) (network eq https://www.googleapis.com/compute/v1/projects/pavithrar-gke-dev/global/networks/e2e-test-pavithrar) (description eq k8s-node-route)'

2) Using "=" for works only for exact match. The following filter returned a 400 "Invalid value for filter ..."
'(name = e2e-test-pavithrar-.*) (network = https://www.googleapis.com/compute/v1/projects/pavithrar-gke-dev/global/networks/e2e-test-pavithrar) (description = k8s-node-route)'

3) Using "=" for exact string match works, but not for regex. The following filter returns the correct matches:
'(network = "https://www.googleapis.com/compute/v1/projects/pavithrar-gke-dev/global/networks/e2e-test-pavithrar") (description = "k8s-node-route")'

4) Using "eq" works for int and bool matches.
a) '(network eq https://www.googleapis.com/compute/v1/projects/pavithrar-gke-dev/global/networks/e2e-test-pavithrar) (description eq k8s-node-route) (priority eq 1000)'
b) '(autoCreateSubnetworks eq false)' for networks api works.

So, for exact string match, the docs are correct. We need to use "=". For regex match, the operator to use is "eq".  "eq" also works for int and bool equality. Finally, "~" is not supported for regex yet.

Reverting this PR will replace all equality filters with "eq", which (based on 1 and 4 above)  should work.

prameshj:
Added a new test and verified that the filters work after the revert(and did not work before)

/assign @bowei

pkg/cloud/filter/filter_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+	projectID = os.Getenv("PROJECT_ID")
+	// test requires e2e cluster to be present in the project specified in order to match these filters
+	routeFilters   = map[string]interface{}{"name": "e2e-test-.*", "description": "k8s-node-route", "priority": 1000}
+	networkFilters = map[string]interface{}{"autoCreateSubnetworks": false}
+)
+
+func TestFilterResources(t *testing.T) {
+	t.Parallel()
+	if projectID == "" {
+		t.Skip("missing projectID... skipping test")
```
rramkumar1: Capitalize

pkg/cloud/filter/filter_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	ctx := context.Background()
+
+	// See https://cloud.google.com/docs/authentication/.
+	// Use GOOGLE_APPLICATION_CREDENTIALS environment variable to specify
+	// a service account key file to authenticate to the API.
+	hc, err := google.DefaultClient(ctx, compute.ComputeScope)
```
rramkumar1: Can just inline context.Background() here since the above var is not used anywhere else.

pkg/cloud/filter/filter_resource_test.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		t.Errorf("Could not get authenticated client: %v", err)
+	}
+
+	svc, err := compute.New(hc)
+	if err != nil {
+		t.Errorf("Could not initialize gke client: %v", err)
```
rramkumar1: You use %v is some places and %s in other. Probably best to stay consistent to one. I would say %v since that is probably used more widely elsewhere.

pkg/cloud/filter/filter_resource_test.go
prameshj: done

bowei:
It would be better if we are keeping the upper layer Regexp() functions to keep the semantic meaning until the translation into GCE API.

prameshj:
> It would be better if we are keeping the upper layer Regexp() functions to keep the semantic meaning until the translation into GCE API.

I just tried making the change so we keep Regexp functions, but use 'eq' operator and use '=' for exact matches. Turns out we cannot do a mix of 2operators in the API query:
'(name eq e2e-test-.*) (description eq k8s-node-route) (priority = 1000)' fails, but '(name eq e2e-test-.*) (description eq k8s-node-route) (priority eq 1000)' works.

We could keep Regexp and equals but fill in the same operator 'eq' always.

pkg/cloud/filter/filter.go
```diff
@@ -199,13 +199,15 @@ func (fp *filterPredicate) String() string {
 	var op string
 	switch fp.op {
 	case regexpEquals:
-		op = "~"
+		op = "eq"
```
bowei: add comment "GCE API maps regexp comparison to `eq`"

PR #25, opened 2019-12-17
During polling, return error on error instead of continued polling

rramkumar1:
Return error on error instead of continuing polling (and likely running into the same error)

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/k8s-cloud-provider/25)

### DIFF
--- a/pkg/cloud/service.go
+++ b/pkg/cloud/service.go
@@ -86,7 +86,7 @@ func (s *Service) pollOperation(ctx context.Context, op operation) error {
 		// returning ctx.Err().
 		select {
 		case <-ctx.Done():
-			klog.V(5).Infof("op.pollOperation(%v, %v) not completed, poll count = %d, ctx.Err = %v", ctx, op, pollCount, ctx.Err())
+      klog.V(5).Infof("op.pollOperation(%v, %v) not completed, poll count = %d, ctx.Err = %v", ctx, op, pollCount, ctx.Err())
 			return ctx.Err()
 		default:
 			// ctx is not canceled, continue immediately
@@ -95,16 +95,13 @@ func (s *Service) pollOperation(ctx context.Context, op operation) error {
 		pollCount++
 		klog.V(5).Infof("op.isDone(%v) waiting; op = %v, poll count = %d", ctx, op, pollCount)
 		s.RateLimiter.Accept(ctx, op.rateLimitKey())
-		done, err := op.isDone(ctx)
-		if err != nil {
+		switch done, err := op.isDone(ctx); {
+		case err != nil:
 			klog.V(5).Infof("op.isDone(%v) error; op = %v, poll count = %d, err = %v, retrying", ctx, op, pollCount, err)
-		}
-
-		if done {
-			break
+			return err
+		case done:
+                        klog.V(5).Infof("op.isDone(%v) complete; op = %v, poll count = %d, op.err = %v", ctx, op, pollCount, op.error())
+			return op.error()
 		}
 	}
-
-	klog.V(5).Infof("op.isDone(%v) complete; op = %v, poll count = %d, op.err = %v", ctx, op, pollCount, op.error())
-	return op.error()
 }
--- a/pkg/cloud/service_test.go
+++ b/pkg/cloud/service_test.go
@@ -22,57 +22,98 @@ import (
 	"testing"
 )
 
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


rramkumar1:
cc @qfel

pkg/cloud/service.go
```diff
@@ -86,7 +86,7 @@ func (s *Service) pollOperation(ctx context.Context, op operation) error {
 		// returning ctx.Err().
 		select {
 		case <-ctx.Done():
-			klog.V(5).Infof("op.pollOperation(%v, %v) not completed, poll count = %d, ctx.Err = %v", ctx, op, pollCount, ctx.Err())
+      klog.V(5).Infof("op.pollOperation(%v, %v) not completed, poll count = %d, ctx.Err = %v", ctx, op, pollCount, ctx.Err())
```
bowei: fix indent?

pkg/cloud/service.go
```diff
@@ -95,16 +95,13 @@ func (s *Service) pollOperation(ctx context.Context, op operation) error {
...
-
-		if done {
-			break
+		switch done, err := op.isDone(ctx); {
+		case err != nil:
+      klog.V(5).Infof("op.isDone(%v) error; op = %v, poll count = %d, err = %v, retrying", ctx, op, pollCount, err)
+			return err
+		case done:
+      klog.V(5).Infof("op.isDone(%v) complete; op = %v, poll count = %d, op.err = %v", ctx, op, pollCount, op.error())
+			return op.error()
```
bowei: missing return value when !done?

pkg/cloud/service.go
```diff
@@ -95,16 +95,13 @@ func (s *Service) pollOperation(ctx context.Context, op operation) error {
...
 		klog.V(5).Infof("op.isDone(%v) waiting; op = %v, poll count = %d", ctx, op, pollCount)
 		s.RateLimiter.Accept(ctx, op.rateLimitKey())
-		done, err := op.isDone(ctx)
-		if err != nil {
-			klog.V(5).Infof("op.isDone(%v) error; op = %v, poll count = %d, err = %v, retrying", ctx, op, pollCount, err)
-		}
-
-		if done {
-			break
+		switch done, err := op.isDone(ctx); {
```
bowei: switch seems extraneous:
```
done, err := op.isDone(ctx)
if err != nil {
  ...
  return ...
}
if done {
  ...
  return ...
}

```

pkg/cloud/service_test.go
```diff
@@ -22,63 +22,104 @@ import (
...
-	// Verify context is cancelled by now.
-	<-ctxCancelled.Done()
-	// Check that pollOperation returns because the context is cancelled.
-	err = s.pollOperation(ctxCancelled, fo)
-	if err == nil {
-		t.Errorf("pollOperation() = nil, want: %v", ctxCancelled.Err())
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			s := Service{RateLimiter: &NopRateLimiter{}}
+			ctx, cfn := context.WithTimeout(context.Background(), 3*time.Second)
```
bowei: we should use 1ms or less (I'm assuming test does not rely on actual timeout)

qfel: The timeout is to have decent failures if the test deadlocks (ie. without the corresponding change in service.go). 1ms looks extremely low, can you really bet there will be enough CPU in testing envs to guarantee everything completes within 1ms? 3s seems much more practical - and it kicks in only if the code deadlocks anyway.

pkg/cloud/service.go
qfel: Seems like a pretty canonical use of switch over a sequence of ifs, eg. https://golang.org/doc/effective_go.html#switch

pkg/cloud/service.go
qfel: What do you mean? This function is intended to poll-loop, in particular the previous version continued on !done

pkg/cloud/service.go
rramkumar1: Done.

pkg/cloud/service.go
rramkumar1: If !done, then it loops again which is expected right?

bowei: ok -- my diff was not in unified so I mismatched the braces

pkg/cloud/service.go
bowei: sgtm

pkg/cloud/service_test.go
bowei: As long as we aren't having unit tests waiting around for 3 s each, sgtm

Issue #29, opened 2020-01-13
GCEInstances.List() doesn't retry


bertinatto:
We've seen some flakes in our CI due to a 503 response from Google API. Example:

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-gcp-4.4/341#1:build-log.txt%3A15854

Looking deeper in the code, it seems like the 503 response comes from a `GCEInstances.List()` call (I'm linking the k8s code because GH doesn't open the `gen.go` file from this repo):

https://github.com/kubernetes/kubernetes/blob/f3df7a2fdb16dd7b00a9d357e337ed48a6f70d45/vendor/github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/gen.go#L13427-L13440

I noticed that `GCEInstances.List()` doesn't retry when an error is returned, however, `GCEInstances.Insert()` does:

https://github.com/kubernetes/kubernetes/blob/f3df7a2fdb16dd7b00a9d357e337ed48a6f70d45/vendor/github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/gen.go#L13484

I'm sure there's a reason for that, but I was wondering if it'd be possible to make `GCEInstances.List()` retry as well, just like other methods of `GCEInstances`.

bowei:
@spencerhance @rramkumar1 please take a look

bertinatto:
ping @bowei @spencerhance @rramkumar1 
Could you take a look at this, please?

spencerhance:
@bertinatto FYI I'm working on a fix for this, since the generated API doesn't export the page token, we just have to implement the paging logic ourselves

spencerhance:
Actually it turns out we can avoid that entirely since the first page token is nil, and we don't care about resetting it back further.

PR #33, opened 2020-02-05
Implement retry logic for List() Operations

spencerhance:
List() calls will retry until successful or the context is done.  PageToken()
is used to avoid re-trying successful pages.

This should resolve https://github.com/GoogleCloudPlatform/k8s-cloud-provider/issues/29

Up for review but hold off on merging since I'm trying to figure out a better way to test this.

### DIFF
--- a/pkg/cloud/gen/main.go
+++ b/pkg/cloud/gen/main.go
@@ -728,15 +728,33 @@ func (g *{{.GCEWrapType}}) List(ctx context.Context, zone string, fl *filter.F) 
 		call.Filter(fl.String())
 	}
 	var all []*{{.FQObjectType}}
+  var nextPageToken string
 	f := func(l *{{.ObjectListType}}) error {
 		klog.V(5).Infof("{{.GCEWrapType}}.List(%v, ..., %v): page %+v", ctx, fl, l)
+		nextPageToken = l.NextPageToken
 		all = append(all, l.Items...)
 		return nil
 	}
-	if err := call.Pages(ctx, f); err != nil {
-		klog.V(4).Infof("{{.GCEWrapType}}.List(%v, ..., %v) = %v, %v", ctx, fl, nil, err)
-		return nil, err
-	}
+	
+  var retryCount int
+  for {
+    select {
+		case <-ctx.Done():
+			klog.V(5).Infof("{{.GCEWrapType}} call.Pages(%v, _) not completed, poll count = %d, ctx.Err = %v", ctx, retryCount, ctx.Err())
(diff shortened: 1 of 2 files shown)


spencerhance:
@bowei Can you review this since you're the most familiar with this code?  Also if you have any suggestions on unit testing that would be appreciated; we don't actually run any of the gen functions in the unit tests.

bertinatto:
@spencerhance, @bowei : Thank you for looking into this!

If you think this is the way to go, I can help testing this patch in our CI for a couple of weeks.

BenTheElder:
[verifying that prow webhooks work]
/woof

bertinatto:
Hi @spencerhance, @bowei. Do you have any thoughts/concerns on this PR?

pkg/cloud/gen/main.go
```diff
@@ -728,15 +728,33 @@ func (g *{{.GCEWrapType}}) List(ctx context.Context, zone string, fl *filter.F)
...
+		case <-ctx.Done():
+			klog.V(5).Infof("{{.GCEWrapType}} call.Pages(%v, _) not completed, poll count = %d, ctx.Err = %v", ctx, retryCount, ctx.Err())
+			return nil, ctx.Err()
+		default:
+			// ctx is not canceled, continue immediately
+		}
+		if nextPageToken != "" {
+			call.PageToken(nextPageToken)
+		}
+		if err := call.Pages(ctx, f); err != nil {
```
bowei, line 751: Are there different kinds of errors we want to handle differently?

spencerhance: I don't think so, are there any specific response codes you think we should handle?  We don't handle response codes individually in the other functions AFAIK

bertinatto:
@bowei @spencerhance do you need any help testing this?

pkg/cloud/gen/main.go
bertinatto: @bowei @spencerhance please let me know if you need any help with this! We're seeing some flakes in our CI that I think would be solved with this patch.

spencerhance:
@bowei Is there anything else I need to do to close this out?

bowei:
Can we address the question about the kinds of errors we would or wouldn't want to retry? Also, I get concerned that current calls to List() that return with an error when called with a no deadline Context will now just hang.

bowei:
Spencer can you update or abandon this PR?

spencerhance:
Closing for now since it's somewhat of a high risk / low reward PR - callers can always just wrap List() in a retry loop.

PR #36, opened 2020-02-21
Update dependency google.golang.org/api

skmatti:
This is required for Global Access GA to be picked up.


rramkumar1:
@skmatti Can you add something in the description indicating why this is needed?

spencerhance:
Is this still required?

skmatti:
I don't think so.

PR #38, opened 2020-04-14
Fix klog.V(x) references for klog v2

dims:
klog.V(x) is no longer a boolean. We need to use the `.Enabled()` method


dims:
/assign @bowei @spencerhance 

cc @tanjunchen @serathius

tanjunchen:
get it

bowei:
This is not quite sufficient -- you need to fix references in the code generator.

bowei:
https://github.com/GoogleCloudPlatform/k8s-cloud-provider/compare/master...bowei:fix-klog?expand=1

dims:
@bowei thanks! pulled in your two commits.

PR #44, opened 2020-07-02
Fix NotRegex filter

jkaniuk:
Small fix for NotRegex filter


bowei:
Can we rebase? Also, would like to see unit test if possible.

jkaniuk:
There is a unit test. Not that I remember the context of this change after 2 years. Where is it used?

bowei:
Looks like this is pretty stale -- please reopen if this is still relevant.

PR #53, opened 2021-05-01
Add Service Attachment Beta APIs

swetharepakula:
This PR also bumps google.golang.api to v0.46.0


swetharepakula:
Currently master fails to build on go1.16. #54 fixes that. This PR has been rebased on #54 and will need to be updated once that merges.

PR #56, opened 2021-05-05
Add support for ListUsable() methods

spencerhance:
In classic fashion, this is kind of hassle because the GCE API returns a different type called `UsableX` (where x is the resource).

Tested on Subnetworks, which is the only object that has a ListUsable method in the API.


pkg/cloud/gen/main.go
```diff
@@ -618,6 +627,44 @@ func (m *{{.MockWrapType}}) AggregatedList(ctx context.Context, fl *filter.F) (m
 }
 {{- end}}
 
+{{- if .ListUsable}}
+// List all of the objects in the mock.
```
bowei, line 631: is this the right comment?

PR #59, opened 2021-06-21
Cover nil for ResourceID Equal()

pacoxu:
See https://github.com/kubernetes/kubernetes/pull/102913#issuecomment-864175076

> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x37d725a]
> 
> A failure in legacy-cloud-providers/gce :/
> This directory is blocked for commits.
> 
> Looks like we need a nil pointer check here and return false if "other" is nil:

### DIFF
--- a/pkg/cloud/utils.go
+++ b/pkg/cloud/utils.go
@@ -39,16 +39,20 @@ type ResourceID struct {
 
 // Equal returns true if two resource IDs are equal.
 func (r *ResourceID) Equal(other *ResourceID) bool {
-	if r.ProjectID != other.ProjectID || r.Resource != other.Resource {
-		return false
+	switch {
+	case r == nil && other == nil:
+	  return true
+	case r == nil || other == nil:
+	  return false
+	case r.ProjectID != other.ProjectID || r.Resource != other.Resource:
+	  return false
+	case r.Key != nil && other.Key != nil:
+	  return *r.Key == *other.Key
+	case r.Key == nil && other.Key == nil:
+	  return true
+	default:
+	  return false
 	}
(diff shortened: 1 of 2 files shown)


pkg/cloud/utils.go
```diff
@@ -39,6 +39,11 @@ type ResourceID struct {
 
 // Equal returns true if two resource IDs are equal.
 func (r *ResourceID) Equal(other *ResourceID) bool {
+	if r == nil && other == nil {
+		return true
+	} else if r == nil || other == nil {
```
bowei, line 44: don't need `else` (return will short circuit)

bowei: you might also want to use switch to make this more clear:

```
switch {
case r == nil && other == nil:
  return true
case r == nil || other == nil:
  return false
case r.ProjectID != other.ProjectID || r.Resource != other.Resource:
  return false
case r.Key != nil && other.Key != nil:
  return *r.Key == *other.Key
case r.Key == nil && other.Key == nil:
  return true
default:
  return false
}
```

bowei: Also -- this needs to be unit tested

bowei:
unit test is missing

pacoxu (commented):
Updated.

pacoxu:
@bowei would you release k8s-cloud-provider a new version?
Then, I can update to a specified version in k/k.

pacoxu:
kindly ping for 1.22 code freeze next week.

PR #62, opened 2021-06-21
Add Subnetwork Patch method and fix service type version

prksu:
This PR adds subnet Patch method and fixes service type version

/assign @bowei


bowei:
Thanks for your contribution

/lgtm
/approve

PR #63, opened 2021-06-24
Add images service

prksu:
This PR adds compute images service


prksu:
@bowei could you please take a look at this PR? thanks!

PR #76, opened 2023-02-17
Introduce ThrottlingStrategy and StrategyRateLimiter backed by strategy

alexkats:
Introduce ThrottlingStrategy and StrategyRateLimiter backed by throttling strategy


pkg/cloud/ratelimit.go
```diff
@@ -76,15 +79,23 @@ func (rl *AcceptRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+func (rl *AcceptRateLimiter) Observe(_ error, _ *RateLimitKey) {
+}
+
 // NopRateLimiter is a rate limiter that performs no rate limiting.
 type NopRateLimiter struct {
 }
 
 // Accept everything immediately.
-func (*NopRateLimiter) Accept(ctx context.Context, key *RateLimitKey) error {
+func (*NopRateLimiter) Accept(_ context.Context, _ *RateLimitKey) error {
```
bowei, line 122: You can just leave out the name (don't need the "_")

pkg/cloud/ratelimit.go
```diff
@@ -76,15 +79,23 @@ func (rl *AcceptRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
 	return nil
 }
 
+// Observe does nothing.
+func (*NopRateLimiter) Observe(_ error, _ *RateLimitKey) {
```
bowei, line 127: Same here 

func (*NopRateLimiter) Observe(error, *RateLimitKey)

pkg/cloud/ratelimit.go
```diff
@@ -45,6 +45,9 @@ type RateLimiter interface {
 	// Accept returns an error if the given context ctx was canceled
 	// while waiting for acceptance into the queue.
 	Accept(ctx context.Context, key *RateLimitKey) error
+	// Observe uses the RateLimitKey to handle response results, which may affect
+	// the sleep time for the Accept function.
+	Observe(err error, key *RateLimitKey)
```
bowei, line 52: Observe(ctx context.Context, err error, key* RateLimitKey)

We should pass the context. This allows for per-call context if the caller wants to customize it that way.

pkg/cloud/ratelimit.go
```diff
@@ -76,15 +79,23 @@ func (rl *AcceptRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
 	return nil
 }
 
+// Observe does nothing.
+func (rl *AcceptRateLimiter) Observe(_ error, _ *RateLimitKey) {
```
bowei, line 76: You can just leave out the name (don't need the "_")

bowei:
First commit LGTM after comments are address. I suggest sending that as a separate PR so you can get it committed and out of the way.

alexkats:
Thanks for the review, I addressed the comments and created #77 with the first commit, this PR goes on top of that changes.

pkg/cloud/ratelimit.go
```diff
@@ -64,25 +66,54 @@ type AcceptRateLimiter struct {
...
 // Observe does nothing.
 func (rl *AcceptRateLimiter) Observe(_ error, _ *RateLimitKey) {
 }
 
+// ThrottlingStrategy handles delays based on feedbacks provided.
+type ThrottlingStrategy interface {
+	// GetDelay returns the delay for next request.
+	GetDelay() time.Duration
+	// PushFeedback recalculates the next delay based on the feedback.
+	PushFeedback(err error)
```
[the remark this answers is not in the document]
alexkats: Agreed, thanks!

pkg/cloud/ratelimit.go
```diff
@@ -64,25 +66,54 @@ type AcceptRateLimiter struct {
...
 }
 
 // Observe does nothing.
 func (rl *AcceptRateLimiter) Observe(_ error, _ *RateLimitKey) {
 }
 
+// ThrottlingStrategy handles delays based on feedbacks provided.
+type ThrottlingStrategy interface {
+	// GetDelay returns the delay for next request.
+	GetDelay() time.Duration
```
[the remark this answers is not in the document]
alexkats: Done

pkg/cloud/ratelimit.go
```diff
@@ -64,25 +66,54 @@ type AcceptRateLimiter struct {
...
+	// PushFeedback recalculates the next delay based on the feedback.
+	PushFeedback(err error)
+}
+
+// StrategyRateLimiter wraps a ThrottlingStrategy with RateLimiter parameters.
+type StrategyRateLimiter struct {
+	// strategy is the underlying throttling strategy.
+	strategy ThrottlingStrategy
+	lock     sync.Mutex
```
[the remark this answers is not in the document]
alexkats: Done

pkg/cloud/ratelimit.go
```diff
@@ -64,25 +66,54 @@ type AcceptRateLimiter struct {
...
+	return &StrategyRateLimiter{
+		strategy: strategy,
+		clock:    clock.RealClock{},
+	}
+}
+
+// Accept block for the delay provided by the strategy or until context.Done(). Key is ignored.
+func (rl *StrategyRateLimiter) Accept(ctx context.Context, _ *RateLimitKey) error {
+	return acceptInternal(ctx, func() {
+		rl.lock.Lock()
```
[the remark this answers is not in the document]
alexkats: Yes, it doesn't make sense, since it'll be blocked anyway, but by delay. Having an explicit lock makes it easier to track required delay for each request, without the lock it should be calculated like some base delay from the strategy + remaining delay from the request(s) in the queue.

pkg/cloud/ratelimit.go
```diff
@@ -76,15 +81,66 @@ func (rl *AcceptRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
 	return nil
 }
 
+// Observe does nothing.
+func (rl *AcceptRateLimiter) Observe(context.Context, error, *RateLimitKey) {
+}
+
+// ThrottlingStrategy handles delays based on feedbacks provided.
```
bowei, line 88: I would like to review ThrottlingStrategy and StrategyRateLimiter separately in a different PR. Can you split this into a separate change?

bowei:
If you split out the new Strategy throttler, I can LGTM this change

alexkats:
Ok, I rebased this PR after merging another one, so it now contains only one commit with throttling strategy changes

bowei:
Did you run `go mod tidy` to remove duplicate versions?

pkg/cloud/ratelimit.go
```diff
@@ -83,6 +85,49 @@ func (rl *AcceptRateLimiter) Accept(ctx context.Context, _ *RateLimitKey) error
...
+func NewStrategyRateLimiter(strategy ThrottlingStrategy) *StrategyRateLimiter {
+	return &StrategyRateLimiter{
+		strategy: strategy,
+		clock:    clock.RealClock{},
+	}
+}
+
+// Accept block for the delay provided by the strategy or until context.Done(). Key is ignored.
+func (rl *StrategyRateLimiter) Accept(ctx context.Context, _ *RateLimitKey) error {
+	rl.lock.Lock()
```
bowei, line 115: It's not clear to me why this lock is being used. What is it guarding?

alexkats: It's hard to tell about what's guarded exactly, it's kind of guarding a combination of `Delay` call to strategy and sleeping. Here's a motivation for it:
The `Delay()` function of a strategy returns the amount of time to wait before sending the request. But the strategy isn't keeping track of requests, so the delay for 2 requests coming at the same time will be the same, meaning we won't get any delay between first and second requests. With the lock we'll wait until the first request is allowed to go, and then the second request can get its delay value.

There's a possibility for a strategy to return not the delay, but the timestamp to wait until (keeping more info inside each strategy implementation), but it will make more time for a recalculated delay to come into effect. Here's an example of a situation with and without the lock:

**Without the lock**:
Current delay for a strategy is 5 seconds.
1. Requests 1-3 come at T=0 and get timestamps for proceeding with the request T=5, T=10, T=15
2. Request 1 starts the execution at T=5 and got the response at T=6, which is observed by the strategy leading to setting the delay to 1 second.
3. Requests 2-3 won't get updated timestamp for proceeding, so the last request will still start at T=15.

**With the lock**:
Current delay for a strategy is 5 seconds.
1. Requests 1-3 come at T=0, request 1 gets delay of 5 seconds with the lock, while requests 2-3 are blocked.
2. Request 1 starts the execution at T=5, request 2 gets delay of 5 seconds (until T=10) with the lock.
3. Request 1 got the response at T=6, which is observed by the strategy leading to setting the delay to 1 second.
4. Request 2 starts the execution at T=10, request 3 gets delay of 1 second (until T=11) with the lock.
5. Request 3 starts the execution at T=11.

alexkats:
> Did you run `go mod tidy` to remove duplicate versions?

Yes, I double-checked now, nothing changed after I tried to run it

pkg/cloud/ratelimit.go
bowei: I think this might need to consider the RateLimitKey, no? With this implementation, you will end up queueing all of the requests, not just ones with a given API. I think you need to maintain queues for each API call?

alexkats: In general yes, but the idea behind was to wrap `StrategyRateLimiter` with some other `RateLimiter` to consider `RateLimitKey` there, similar to `AcceptRateLimiter`. It makes sense though to add an explicit comment about the usage of such a rate limiter. Wdyt?

bowei:
As discussed, I think this change makes more sense in the repo that uses this, due to the need to add copious metrics and logging. Closing.

PR #78, opened 2023-02-28
Add utils functions for RelativeResourceNameWithVersion

spencerhance:
This func generates a string very similar to a SelfLink but without including the "googleapis.com" endpoint."


bowei:
Let's discuss -- how are we going to use the new function? Don't we need to make the http://googleapis.com... configurable for all instances?

spencerhance:
Closing in favor of https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/79

PR #79, opened 2023-03-03
Change gce endpoint const values into exported vars.

spencerhance:
This will allow packages who import this to configure the endpoints. These endpoints are only used for SelfLink().


pkg/cloud/utils.go
```diff
@@ -24,10 +24,10 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 )
 
-const (
-	gaPrefix    = "https://www.googleapis.com/compute/v1"
-	alphaPrefix = "https://www.googleapis.com/compute/alpha"
-	betaPrefix  = "https://www.googleapis.com/compute/beta"
+var (
```
bowei, line 27: Would prefer to not export the variables directly.

var (
  gaPrefix = ..

)

func SetAPIDomain(domain string) {
  gaPrefix = doamin + "/compute/v1"
  ...
}

spencerhance: Updated

pkg/cloud/utils.go
```diff
@@ -24,12 +24,18 @@ import (
...
 )
 
-const (
+var (
 	gaPrefix    = "https://www.googleapis.com/compute/v1"
 	alphaPrefix = "https://www.googleapis.com/compute/alpha"
 	betaPrefix  = "https://www.googleapis.com/compute/beta"
 )
 
+func SetAPIDomain(domain string) {
```
bowei, line 33: Doc string is missing

bowei: // SetAPIDomain sets the root of the URL for the API. The default domain is
// "https://www.googleapis.com".

spencerhance: Whoops, just fixed.

bowei:
add a unit test that checks the SetDomain took effect in the call to the API and this lgtm

spencerhance:
> add a unit test that checks the SetDomain took effect in the call to the API and this lgtm

Added a new test, lmk if that wasn't what you meant.

PR #88, opened 2023-03-24
Refactor code generator.

briantkennedy:
* add helper to allow usage with klog v1
* move import paths to constants


briantkennedy:
Fixed build issues

PR #89, opened 2023-03-25
Update google.golang.org/api version to get NEG ipv6 fields

gauravkghildiyal:


gauravkghildiyal:
There's some test failures upstream. Will wait for them to get fixed.

go.mod
```diff
@@ -30,6 +30,6 @@ replace (
 	golang.org/x/net => golang.org/x/net v0.0.0-20210503060351-7fd8e65b6420
```
aojea, line 31: you should not need to use these replace block, replace is for pinning dependencies

aojea: can you try to remove the entire `replace `block and run `go mod tidy && go mod vendor ` again?

gauravkghildiyal:
Thanks for looking into this Antonio!

I tried running the tests locally and they pass for me without any additional changes.

Why I believed it was the older go version was because the error messages are complaining about objects not found in the go standard library: 
image

Looks like these were initially available in the `ioutil` package but then deprecated and moved to the `io` package in go1.16. We're using kubekins-e2e image from 2019 and looks like go1.16 wasn't released then: 
image

go.mod
gauravkghildiyal: _>you should not need to use these replace block, replace is for pinning dependencies_
While I'm not aware of the context of why that was done, but I did see a commit from @spencerhance  to explicitly pin them. Spencer, can you clarify if this is still needed?

_>can you try to remove the entire replace block and run go mod tidy && go mod vendor  again?_
Gave this a shot. It failed with another reason. I believe the explanation in https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/89#issuecomment-1485687732 should be the cause for this too.

aojea:
> Looks like these were initially available in the `ioutil` package but then deprecated and moved to the `io` package in go1.16. We're using kubekins-e2e image from 2019 and looks like go1.16 wasn't released then:

:/

go.mod
spencerhance: We probably don't need to pin dependencies anymore.  We originally did this when we first migrated to go modules to make sure that dependencies wouldn't get updated.

gauravkghildiyal: Thanks Spencer! I removed the pinned dependencies.

bowei:
Gaurav -- can you update and rerun to regenerate the changed files.

gauravkghildiyal:
@bowei 
> Gaurav -- can you update and rerun to regenerate the changed files.

Done

aojea:
LGTM

PR #93, opened 2023-04-18
Adapter for handling structural-based versioning

bowei:
Adds an adapter for easily dealing with the API versioning types.

see doc.go for details on how this is intended to be used.

Review sequence:

* Read doc.go
* visit.go, visit_test.go -- these should be easily understood (will likely split this into its own PR)
* fill.go is just a helper for testing
* copy.go is the main machinery.
* obj.go is what people will end up using.

### DIFF
--- a/pkg/cloud/api/copy.go
+++ b/pkg/cloud/api/copy.go
@@ -1,0 +1,317 @@
+/*
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package api
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+)
+
+type missingFieldOnCopy struct {
+	Path  Path
+	Value any
+}
+
+// copierOption are options that customize the behavior of the internal copier.
+type copierOption func(*copier)
+
+// copierLogS considers logging for the copier to use the given func f.
+func copierLogS(f func(msg string, kv ...any)) copierOption {
+	return func(c *copier) { c.logSFn = f }
+}
+
+func newCopier(opts ...copierOption) *copier {
+	c := &copier{}
+	for _, o := range opts {
+		o(c)
+	}
+	return c
+}
+
+type copier struct {
+	// logSFn is an optional structured log function, matching the
+	// signature from klog/v2.
+	logSFn func(msg string, kv ...any)
+
+	missing []missingFieldOnCopy
+}
+
+func (c *copier) logS(msg string, kv ...any) {
+	if c.logSFn == nil {
+		return
+	}
+	c.logSFn(msg, kv...)
+}
+
+func (c *copier) do(dest, src reflect.Value) error {
+	return c.doValues(Path{}, dest, src)
+}
+
+func (c *copier) doValues(p Path, dest, src reflect.Value) error {
+	switch {
+	case isBasicV(dest) && isBasicV(src):
+		return c.doBasic(p, dest, src)
+	case src.Type().Kind() == reflect.Pointer && dest.Type().Kind() == reflect.Pointer:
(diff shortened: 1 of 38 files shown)


pkg/cloud/api/copy.go
```diff
@@ -0,0 +1,303 @@
...
+	c.logSFn(msg, kv...)
+}
+
+func copyValues(dest, src reflect.Value) (*copyContext, error) {
+	cc := &copyContext{}
+	err := copyValuesImpl(cc, Path{}, dest, src)
+	return cc, err
+}
+
+func copyValuesImpl(cc *copyContext, p Path, dest, src reflect.Value) error {
```
briantkennedy, line 50: nit: the `cc *copyContext` on all the methods tells me these should probably be struct member functions

pkg/cloud/api/copy.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+			continue
+		}
+
+		// ServerResponse should be skipped.
+		if (p.Equal(Path{}) || p.Equal(Path{}.Pointer())) && fieldName == "ServerResponse" {
+			continue
+		}
+
+		if fieldName == "NullFields" || fieldName == "ForceSendFields" {
```
briantkennedy, line 158: do we need to associate these fields with a type?

pkg/cloud/api/obj.go
```diff
@@ -0,0 +1,300 @@
...
+		return cycleCheck(p.MapIndex("x"), t.Elem(), seen)
+	case reflect.Struct:
+		typeName := fmt.Sprintf("%s/%s", t.PkgPath(), t.Name())
+		for _, seenTypeName := range seen {
+			if typeName == seenTypeName {
+				return fmt.Errorf("recursive type found at %s: %s", p, typeName)
+			}
+		}
+		// Add this struct type to the list of types seen on this path.
+		seen = append(seen, fmt.Sprintf("%s/%s", t.PkgPath(), t.Name()))
```
briantkennedy, line 96: pop the stack after processing?

pkg/cloud/api/path.go
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+package api
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Path specifies a field in nested object. The type of the reference
```
briantkennedy, line 24: can we make this more json-path like for the slice index and maybe map key?

pkg/cloud/api/diff.go
```diff
@@ -0,0 +1,71 @@
...
+	// include in API requests with the JSON null value. By default, fields
+	// with empty values are omitted from API requests. However, any field
+	// with an empty value appearing in NullFields will be sent to the
+	// server as null. It is an error if a field in this list has a
+	// non-empty value. This may be used to include null fields in Patch
+	// requests.
+
+*/
+
+type DiffField struct {
```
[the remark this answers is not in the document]
bowei: see above.

pkg/cloud/api/obj.go
```diff
@@ -0,0 +1,202 @@
...
+
+type conversionErrors struct {
+	missingFields []missingFieldOnCopy
+}
+
+// VersionedObject wraps the standard GA, Alpha, Beta versions of a GCP
+// resource. By accessing the object using Edit(), EditAlpha(). EditBeta(),
+// VersionedObject will ensure that common fields between the versions of the
+// object are in sync.
+type VersionedObject[GA any, Alpha any, Beta any] struct {
```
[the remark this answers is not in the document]
bowei: For resources we know about yes. We can add some handling for a dummy type if there is such a use case...

pkg/cloud/api/copy.go
bowei: true -- was debating this, I can flip if you think this works better.

pkg/cloud/api/copy.go
bowei: Are you saying we should also check the type in case there is a field with the name but it's not actually a metafield?

As far as I can tell -- these are always generated in every struct level.

pkg/cloud/api/obj.go
bowei: Is this needed? seen gets thrown away when the function returns...

pkg/cloud/api/path.go
bowei: the single character makes it easy to parse for someone using the path. Can you give an example of what you are thinking?

briantkennedy:
lgtm

briantkennedy (approved):
lgtm

PR #112, opened 2023-05-09
Introduce a package-private type contextKey.

krzykwas:
Using &struct{}{} as a context key has a risk of running into collisions. Besides that enhance the logging instructions so they can better communicate the exact type that was extracted from context.

### DIFF
--- a/pkg/cloud/observe.go
+++ b/pkg/cloud/observe.go
@@ -32,7 +32,9 @@ type CallObserver interface {
 	End(ctx context.Context, key *RateLimitKey, err error)
 }
 
-var callObserverContextKey = &struct{}{}
+type contextKey string
+
+var callObserverContextKey = contextKey("call observer")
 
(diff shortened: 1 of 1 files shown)


bowei:
What is the collision and why does string fix this?

krzykwas:
&struct{}{} is not protected and could be reused by some other code putting something into context. Using a type scoped to the package instead makes an overwrite possible only within the same package.

`type contextKey string` allows specifying an additional string key, so a collision can only happen if the same string is used twice in the same package.

krzykwas:
You mean `type contextKey struct{}{}`. Yes, but there can only be one such key in a package.

briantkennedy:
> What is the collision and why does string fix this?

This is the recommended way to use context key: https://pkg.go.dev/context#WithValue

briantkennedy (approved):

PR #116, opened 2023-05-15
Add RateLimiter.Observe to some missing paths

alexkats:
* Add `RateLimiter.Observe` to some missing paths
* Make `RateLimiter` tests parallel

### DIFF
--- a/pkg/cloud/gen/main.go
+++ b/pkg/cloud/gen/main.go
@@ -985,6 +985,8 @@ func (g *{{.GCEWrapType}}) AggregatedList(ctx context.Context, fl *filter.F) (ma
 		klog.V(4).Infof("{{.GCEWrapType}}.AggregatedList(%v, %v) = %v, %v", ctx, fl, nil, err)
 		return nil, err
 	}
+	g.s.RateLimiter.Observe(ctx, nil, ck)
+
 	if kLogEnabled(4) {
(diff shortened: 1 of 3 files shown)


pkg/cloud/gen/main.go
```diff
@@ -985,6 +985,8 @@ func (g *{{.GCEWrapType}}) AggregatedList(ctx context.Context, fl *filter.F) (ma
 		klog.V(4).Infof("{{.GCEWrapType}}.AggregatedList(%v, %v) = %v, %v", ctx, fl, nil, err)
 		return nil, err
 	}
+	g.s.RateLimiter.Observe(ctx, nil, ck)
```
bowei, line 988: Is this path missing callObserverEnd as well?

pkg/cloud/gen/main.go
```diff
@@ -1032,6 +1034,7 @@ func (g *{{.GCEWrapType}}) ListUsable(ctx context.Context, fl *filter.F) ([]*{{.
 		klog.V(4).Infof("{{.GCEWrapType}}.ListUsable(%v, ..., %v) = %v, %v", ctx, fl, nil, err)
 		return nil, err
 	}
+	g.s.RateLimiter.Observe(ctx, nil, ck)
```
bowei, line 1037: callObserverEnd?

pkg/cloud/gen/main.go
alexkats: Right, thanks for pointing out, fixed

PR #119, opened 2023-05-19
Rename FrozenResource -> Resource, Resource -> MutableResource

bowei:
These names are a bit more intuitive


pkg/cloud/api/mutable_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	SetAlpha(src *Alpha) error
+	// SetBeta the value to src. This skips some of the field
+	// validation in Access* so should only be used with a valid
+	// object returned from GCE.
+	SetBeta(src *Beta) error
+
+	// Freeze the resource to a read-only copy. It is an error if it is ambiguous
+	// which version is the correct one i.e. not all fields can be represented in a
+	// single version of the resource.
+	Freeze() (Resource[GA, Alpha, Beta], error)
```
robscott, line 155: Potentially confusing to have a `MutableResource` that you can freeze (and therefore it would no longer be mutable).

pkg/cloud/api/mutable_resource.go
```diff
@@ -0,0 +1,405 @@
...
+			errors:     &u.errors[BetaToGAConversion],
+		})
+		conversions = append(conversions, convert{
+			dest:       reflect.ValueOf(&u.alpha),
+			copyHelper: func() error { return u.typeTrait.CopyHelperBetaToAlpha(&u.alpha, &u.beta) },
+			errors:     &u.errors[BetaToAlphaConversion],
+		})
+	}
+
+	if flags&postAccessSkipValidation == 0 {
```
robscott, line 241:
```suggestion
	if flags & postAccessSkipValidation == 0 {
```

robscott (commented):
Thanks @bowei! I'm not that familiar with this codebase yet, but have a couple comments.

pkg/cloud/api/mutable_resource.go
bowei: seems like a diff between gofmt versions...

pkg/cloud/api/mutable_resource.go
bowei: This is kind of the point. You first are allowed to mutate it, then you Freeze() to remove the possibility of mutation. This is the lifecycle that is desired.

briantkennedy (approved):

PR #121, opened 2023-05-25
Make SerialExecutor bail on bad arguments earlier

bowei:
- Bail in the constructor of SerialExecutor
- Shortens some stuttering names


briantkennedy (commented):
LGTM

PR #124, opened 2023-06-10
Adding Support for NetworkServices APIs

robscott:
All tests are passing, though we have follow up AIs to add more test coverage for these new APIs. The most significant changes in terms of backwards compatibility and deprecations have already been handled by #130. Now that that PR has merged, this should be relatively straightforward. The vast majority of code changes left in this PR are related to codegen and importing the networkservices package.


bowei:
Can we split this commit into one that does the ResourceID changes, get that resolved and then the codegen for the import? We should try to make the large, mostly boilerplate commit review to be as painless as possible.

bowei:
We should track down the usage of RelativeResourcePath and ResourcePath. They were added quite recently I think to support changing the googleapis.com domain and it may be to possible to either remove these or at least make then non-exported.

@spencerhance

spencerhance:
> They were added quite recently I think to support changing the googleapis.com domain and it may be to possible to either remove these or at least make then non-exported.

RelativeResourcePath() and ResourcePath() appear to have been around since the initial import from k/k 5+ years ago.

The configurable domain change was just to add SetAPIDomain() which is only used by SelfLink().

Looking at [ingress-gce](https://github.com/search?q=repo%3Akubernetes%2Fingress-gce+ResourcePath&type=code) , there only appears to be a couple uses where it's mainly for convenience (you don't have to specify more info than the GCE api needs to find the resource).

I suspect that all downstream users could find a way to switch to SelfLink() instead if we deprecate those two funcs.

robscott:
This PR has been rebased now that #130 has merged and should be significantly simpler to review now. PTAL.

Also worth considering if we want to cut a new release with just #130 before this one merges.

/cc @bowei @spencerhance @swetharepakula

PR #125, opened 2023-06-19
Wrap usage of []Event with EventList

bowei:
- EventList implements order-indep Equal()
- Clarify semantics in the docstrings

### DIFF
--- a/pkg/cloud/rgraph/exec/event.go
+++ b/pkg/cloud/rgraph/exec/event.go
@@ -27,8 +27,30 @@ import (
 type Event interface {
 	// Equal returns true if this event is == to other.
 	Equal(other Event) bool
-	// String implements Stringer.
+	// String implements Stringer. The values returned by String() must satisfy
+	// the same behavior as Equal(), i.e. a.String() == b.String() if and only
+	// if a.Equal(b).
 	String() string
+}
+
+// EventList is a list of events.
+type EventList []Event
+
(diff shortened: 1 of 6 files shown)


pkg/cloud/rgraph/exec/event.go
```diff
@@ -27,10 +27,32 @@ import (
 type Event interface {
 	// Equal returns true if this event is == to other.
 	Equal(other Event) bool
-	// String implements Stringer.
+	// String implements Stringer. The values returned by String() must satisfy
```
briantkennedy, line 30: This contract is going to be a bit difficult to enforce.  I looked at the current impl, and the `%v` used in the `String()` methods is going to format contained pointers as hex-addresses rather than emitting the contained fields.  Can we update `EventList.Equal` or other code / test locations to defend against implementations that violate this constraint?

bowei: Not sure how to fix this? Are you suggesting using some form of reflect.DeepEqual

briantkennedy: Might be able to detect it by keeping a map of `Event.String()` to `Event` then calling `Equal()` on collision.  In any case, the existing events need to get fixed up for this to work.

bowei: To be specific -- you mean in the EventList Equal() method?

PR #128, opened 2023-06-19
Add algo/localplan

bowei:
Adds algo/local_plan

Local plan computes a local plan of action for each pair of nodes in the (got, want) graph pair.

### DIFF
--- a/pkg/cloud/rgraph/algo/localplan/local_plan.go
+++ b/pkg/cloud/rgraph/algo/localplan/local_plan.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


briantkennedy:
This has some already added files listed as add, can you rebase?

bowei:
Fixed -- thanks!

bowei:
Ok, this should be ready now.

pkg/cloud/rgraph/algo/localplan/local_plan.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode"
+)
+
+// Do computes a plan local to each Node in the graph and puts the resulting
+// plan in the "want" Graph. It is required that got and want have the same set
+// of Nodes; Nodes that don't exist need to be marked as with
+// NodeStateDoesNotExist.
+func Do(got, want *rgraph.Graph) error {
```
briantkennedy, line 30: The input/output var won't be obvious at the callsite when reading, can we call this `PopulateWant` or something that conveys we are updating the input and which input is getting modified?  Same idea for `Plan` where it's not quite obvious.

briantkennedy (approved):

PR #129, opened 2023-06-21
Export Fake types; Add EZ graph builder

bowei:
* Expose Fake types beyond package for use for testing
* Adds ez graph builder, a terse way to create resource graphs
* Add ez graph factory for fake resources


pkg/cloud/rgraph/ez/ez.go
```diff
@@ -0,0 +1,146 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package ez is a utility to create complex resource graphs for testing from a
+// concise description by use of naming conventions and default values.
+//
+// This package should only be used for testing. There is little error handling;
+// most errors will result in a panic() to reduce the verbosity of the code.
+package ez
```
briantkennedy, line 22: Can we update the package name to `eztesting` or `graphtesting` since not using this in prod requires people to read the docstring?

pkg/cloud/rgraph/ez/ez.go
```diff
@@ -0,0 +1,146 @@
...
+	// PanicOnAccessErr causes panic() if Accces() returns an error. Otherwise,
+	// the error is ignored.
+	PanicOnAccessErr = 1 << iota
+)
+
+func panicf(s string, args ...any) { panic(fmt.Sprintf(s, args...)) }
+
+type Graph struct {
+	Nodes   []Node
+	Project string
```
robscott, line 44: Why is Project an attribute of both `Graph` and `Node`? (I think it makes sense on Node if/when we need to support cross-project graphs, but I'm not sure why we need this in both places and how we should interpret conflicting values).

pkg/cloud/rgraph/ez/ez.go
briantkennedy: maybe put it at `pkg/cloud/rgraph/testing/ez`?

bowei: ok

pkg/cloud/rgraph/ez/ez.go
bowei: It's to make the common case more terse:

```
		ezg := ez.Graph{
      Project: "proj-1",
			Nodes: []ez.Node{
				{Name: "addr1", Project: "proj-2"},
				{Name: "fr1", Refs: []ez.Ref{{Field: "IPAddress", To: "addr1"}, {Field: "Target", To: "thp"}}},
				{Name: "thp", Refs: []ez.Ref{{Field: "UrlMap", To: "um"}}},
				{Name: "um", Refs: []ez.Ref{{Field: "DefaultService", To: "bs"}}},
				{Name: "bs", Refs: []ez.Ref{{Field: "Backends.Group", To: "us-central1-b/neg1"}, {Field: "Healthchecks", To: "hc"}}},
				{Name: "hc"},
				{Name: "neg1", Zone: "us-central1-b"},

				{Name: "fr2", Options: ez.DoesNotExist},
				{Name: "fr3", Options: ez.DoesNotExist},
			},
		}
```

robscott: So this is essentially a default for Nodes when they're unspecified? If so, maybe a comment would help explain that for someone new to this.

bowei: good point, adding.

pkg/cloud/rgraph/ez/ez.go
bowei: Is `testing` or `test`. one of these is somewhat special?

briantkennedy: I don't believe path the paths are special, but `testing` seems like a common directory or package name for test libraries / helpers.

briantkennedy:
This is LGTM from my side @robscott are your comments resolved?

pkg/cloud/rgraph/testing/ez/ez.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Graph is an easy way to build resource Graphs. This is intended for static
+// testing case construction. Errors are handled by panic().
+type Graph struct {
+	// Nodes in the Graph.
+	Nodes []Node
+	// Project to use by default if not specified in the Node.
+	Project string
+	// Options to use by default in the Graph and Nodes.
+	Options GraphOption
```
robscott, line 50: How are multiple options intended to be represented here?

robscott (commented):
One more question but otherwise LGTM, thanks @bowei!

pkg/cloud/rgraph/testing/ez/ez.go
bowei: bit-wise OR. Let me update the docs

robscott: It looks like you only added a comment on `NodeOption` but not `GraphOption`

pkg/cloud/rgraph/testing/ez/ez.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph"
+)
+
+type GraphOption int
```
robscott, line 32: A comment here that matches what you added to NodeOption would be helpful

bowei: done

robscott:
Thanks!

/lgtm

PR #130, opened 2023-06-22
Adding concept of APIGroup to Services

robscott:
In preparation for adding the NetworkServices API Group, this makes a number of breaking changes:

1. APIGroup is required in ResourceID
2. APIGroup is required in SelfLink
3. RelativeResourceName and ResourcePath are deprecated in favor of SelfLink as they can't support APIGroup

Note: This is a subset of #124 and should include all the breaking changes needed for this transition.


pkg/cloud/utils.go
```diff
@@ -128,9 +139,25 @@ func (r *ResourceID) String() string {
 //	[https://www.googleapis.com/compute/<ver>]/projects/<proj>/global/<res>/<name>
```
bowei, line 139: We should update the docstring

pkg/cloud/utils.go
```diff
@@ -19,30 +19,32 @@ package cloud
...
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 )
 
 var (
-	gaPrefix    = "https://www.googleapis.com/compute/v1"
-	alphaPrefix = "https://www.googleapis.com/compute/alpha"
-	betaPrefix  = "https://www.googleapis.com/compute/beta"
+	computePrefix         = "https://www.googleapis.com/compute"
+	networkServicesPrefix = "https://www.googleapis.com/networkservices"
+	apiGroupRegex         = regexp.MustCompile(`([a-z0-1]*)\/([a-z0-1]*)?\/projects`)
```
bowei, line 31: can we put this closer to the use and document what this is supposed to match

pkg/cloud/utils.go
```diff
@@ -128,9 +139,25 @@ func (r *ResourceID) String() string {
 //	[https://www.googleapis.com/compute/<ver>]/projects/<proj>/global/<res>/<name>
 //	[https://www.googleapis.com/compute/<ver>]/projects/<proj>/regions/<region>/<res>/<name>
 //	[https://www.googleapis.com/compute/<ver>]/projects/<proj>/zones/<zone>/<res>/<name>
+//
+// Note that ParseResourceURL can't round trip partial paths that do not
+// include an API Group.
 func ParseResourceURL(url string) (*ResourceID, error) {
 	errNotValid := fmt.Errorf("%q is not a valid resource URL", url)
 
+	matches := apiGroupRegex.FindStringSubmatch(url)
```
bowei, line 148: Can we make this more precise instead of finding arbitrary submatches? Ideally, we strip off the constant prefix and match exact from start to end of string.

pkg/cloud/utils.go
```diff
@@ -233,21 +262,32 @@ func RelativeResourceName(project, resource string, key *meta.Key) string {
...
 
 // SelfLink returns the self link URL for the given object.
-func SelfLink(ver meta.Version, project, resource string, key *meta.Key) string {
+func SelfLink(apiGroup meta.APIGroup, ver meta.Version, project, resource string, key *meta.Key) string {
 	var prefix string
+
+	switch apiGroup {
+	case meta.APIGroupNetworkServices:
+		prefix = networkServicesPrefix
+	default:
```
bowei, line 271: Can we make this an explicit match to empty?

pkg/cloud/utils_test.go
```diff
@@ -31,16 +31,16 @@ func TestEqualResourceID(t *testing.T) {
 		b *ResourceID
 	}{
 		{
-			a: &ResourceID{"some-gce-project", "projects", nil},
-			b: &ResourceID{"some-gce-project", "projects", nil},
+			a: &ResourceID{"some-gce-project", meta.APIGroupCompute, "projects", nil},
```
bowei, line 34: Seems like we don't test for the network services compute prefix?

pkg/cloud/utils.go
robscott: Unfortunately this function supports a _very_ broad range of inputs and there is no shared prefix we could rely on here:

https://github.com/GoogleCloudPlatform/k8s-cloud-provider/blob/ade9202145be71e04bd11fd952a7829d1fab7c5d/pkg/cloud/utils.go#L119-L130

robscott: This is particularly relevant because we allow the `googleapis.com` domain to be overridden.

pkg/cloud/utils.go
robscott: This is in the same file it's used in, I think including this in the function itself would add a non-trivial performance cost, do you have any other recommendations?

pkg/cloud/utils.go
robscott: This function does not currently return an error. Since we're already updating the function signature to include API Group we could theoretically also return an error, but I think that would make updating projects that depend on this especially painful. We could also do what we're doing elsewhere in this function and add something like `invalid-version` to the returned value, but that does not seem particularly helpful. Any preference here?

pkg/cloud/utils.go
robscott: Updated all of these to have `<apigroup>` instead of `compute`

pkg/cloud/utils_test.go
robscott: Good call - added test for NetworkServices

pkg/cloud/utils.go
bowei: I meant move the definition right above the function that is using it. You can keep at as a static variable.

pkg/cloud/utils.go
bowei: That doesn't mean we can't make the regex more specific...

pkg/cloud/utils.go
bowei: invalid-xxx will be logged so we can potentially see what happened. Given this is very rare, we would still like it to be detectable.

pkg/cloud/utils.go
robscott: As far as I can tell, we can't do anything to predict what comes before API Group because it's configurable via `SetAPIDomain`. We do know that `/projects` will _always_ be after it, and usually a `<version>` path segment will come between API Group and `/projects`. That's what the RegEx is matching for. Any suggestions for how we could make this more specific?

pkg/cloud/utils.go
robscott: Updated, thanks!

pkg/cloud/utils.go
robscott: Updated regex to only match known set of versions and added more test cases, let me know if you'd rather revert to something a bit looser here.

bowei: We should put validation in SetDomain to restrict the domain of that.

pkg/cloud/utils.go
robscott: Updated, thanks1

robscott:
Latest updates:

* Updated `SelfLink()` function so behavior should be unchanged by this PR.
* Added new `SelfLinkWithGroup()` function that accepts API Group as first argument.
* Deprecated `SelfLink()` function in favor of `SelfLinkWithGroup()` function.

If the deprecation of `SelfLink()` causes linting issues if dependent projects choose not to move to `SelfLinkWithGroup()`, we can easily revert that.

pkg/cloud/utils.go
```diff
@@ -93,29 +99,47 @@ func (r *ResourceID) MapKey() ResourceMapKey {
...
 }
 
 // ResourcePath returns the resource path representing this ResourceID.
+// Deprecated: Use SelfLinkWithGroup instead
 func (r *ResourceID) ResourcePath() string {
 	return ResourcePath(r.Resource, r.Key)
 }
 
+// SelfLink returns a URL representing the resource and assumes Compute API Group.
+// Deprecated: Use SelfLinkWithGroup instead
```
bowei, line 114: This is pretty confusing to me. Why do we need to change the contract for a ResourceID? The ResourceID . SelfLink() is already a fully qualified URL including the API Group.

Why can't we have this method treat an empty API group as compute API?

I think we need to discuss this change as this approach feels wrong.

robscott: Yep, that's fair, I was trying to mirror the other functions but you're right that it's unnecessary here. I've reworked this and added new tests.

robscott:
@bowei Updated PR and addressed feedback, PTAL.

PR #135, opened 2023-08-10
Adds parallel queue

bowei:
Parallel queue enables parallel execution with a configurable number of parallel goroutines. Supports context Cancel.

### DIFF
--- a/pkg/cloud/rgraph/algo/parallel_queue.go
+++ b/pkg/cloud/rgraph/algo/parallel_queue.go
@@ -1,0 +1,241 @@
+/*
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package algo
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+type config struct {
+	workerCount int
+	tracer      Tracer
+}
+
+type QueueOption func(*config)
+
+func WorkerCount(n int) QueueOption  { return func(c *config) { c.workerCount = n } }
+func UseTracer(t Tracer) QueueOption { return func(c *config) { c.tracer = t } }
+
+type Tracer func(RunInfo)
+
+// RunInfo records the details of a task.
+type RunInfo struct {
+	// ID for the execution.
+	ID string
+	// Queued is the timestamp of when the task was queued.
+	Queued time.Time
+	// Start is the timestamp of when the task was started as a goroutine.
+	Start time.Time
+	// End is when the task was finished.
+	End time.Time
+	// Err is the result of the task.
+	Err error
+}
+
+// NewParallelQueue returns a new queue instance.
+func NewParallelQueue[T fmt.Stringer](opts ...QueueOption) *ParallelQueue[T] {
+	cfg := config{
+		workerCount: 2,
+		tracer:      func(RunInfo) {},
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+	return &ParallelQueue[T]{
+		c:    cfg,
+		in:   make(chan struct{}, 100),
+		done: make(chan RunInfo, 100),
+	}
+}
+
+// ParallelQueue executes parallel tasks on separate goroutines.
+type ParallelQueue[T fmt.Stringer] struct {
+	c config
+
+	lock    sync.Mutex
+	state   queueState
+	pending []queueElement[T]
+	in      chan struct{}
+	done    chan RunInfo
+	active  int
+}
+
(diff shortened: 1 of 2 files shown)


pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+		workerCount: 2,
+		tracer:      func(RunInfo) {},
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+	return &ParallelQueue[T]{
+		c:    cfg,
+		in:   make(chan struct{}, 100),
+		done: make(chan RunInfo, 100),
```
aojea, line 66: why 100?

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+
+type queueElement[T fmt.Stringer] struct {
+	ri   RunInfo
+	item T
+}
+
+type queueState int
+
+const (
+	stateNotStarted queueState = iota
```
aojea, line 90: nit, I find using the atomic https://pkg.go.dev/sync/atomic values or ints easier to read for these things (and less prone to deadlock errors), instead of using a mutex to protect the `q.state` you just update the atomic type

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+				q.state = stateDone
+				klog.V(2).Infof("Task error, exiting early (pending: %d active: %d): %v", len(q.pending), q.active, ri.Err)
+				q.lock.Unlock()
+				return ri.Err
+			}
+
+			q.lock.Unlock()
+		case <-q.in:
+			klog.V(4).Info("<-q.in")
+			// Fall through to launch().
```
aojea, line 173: but if the state has changed before launch() this element will not be processed because you have a check before launch(), is that the intended behavior?

bowei: It sufficient to have it get kicked from sleep in `select`.

It's fine as the state => stateDone means we are not launching more tasks anyway.

bowei: Let me update the comment to be more specific.

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: I have 10 fingers.

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: Maybe -- but we are grabbing the lock anyway, so for me, it's simpler than reasoning about atomic + lock.

aojea: fair enough

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+		done: make(chan RunInfo, 100),
+	}
+}
+
+// ParallelQueue executes parallel tasks on separate goroutines.
+type ParallelQueue[T fmt.Stringer] struct {
+	c config
+
+	lock    sync.Mutex
+	state   queueState
```
briantkennedy, line 75: please add comments for members

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+// as this can result in a deadlock.
+//
+// Queue execution will stop if op() returns an error or the context
+// is canceled. When an error occurs, there may be goroutines that are
+// continuing to execute. Use q.WaitForOrphans() to wait for them
+// remaining goroutines.
+func (q *ParallelQueue[T]) Run(ctx context.Context, op func(context.Context, T) error) error {
+	q.lock.Lock()
+	if q.state != stateNotStarted {
+		return fmt.Errorf("Run() can only be called once (state=%d)", q.state)
```
briantkennedy, line 133: release lock?

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -0,0 +1,241 @@
...
+// WaitForOrphans will block until remaining op() goroutines
+// finish. Call this if Run() returns an error and you need to know
+// that all remaining threads of execution are done.
+//
+// WaitForOrphans will exit early if ctx is cancelled. This `ctx`
+// should be different from the `ctx` given to Run().
+func (q *ParallelQueue[T]) WaitForOrphans(ctx context.Context) error {
+	q.lock.Lock()
+	if q.state != stateDone {
+		return fmt.Errorf("WaitForOrphans called when not done (state = %d)", q.state)
```
briantkennedy, line 215: unlock

briantkennedy (approved):

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: fixing -- the change merged, so this will be a follow up PR.

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: fixing -- the change merged, so this will be a follow up PR.

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: acknowledged

PR #141, opened 2023-10-18
Add unit test for the generated method code

bowei:
* Added missing type checking for clearer error message in the Pages() method code.

### DIFF
--- a/pkg/cloud/meta/method.go
+++ b/pkg/cloud/meta/method.go
@@ -29,7 +29,7 @@ func newArg(t reflect.Type) *arg {
 Loop:
 	for {
 		switch t.Kind() {
-		case reflect.Ptr:
+		case reflect.Pointer:
 			ret.numPtr++
 			t = t.Elem()
 		default:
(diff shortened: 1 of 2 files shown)


pkg/cloud/meta/method.go
```diff
@@ -234,6 +234,10 @@ func (m *Method) init() {
 			if itemsType.Kind() != reflect.Slice && itemsType.Elem().Kind() != reflect.Ptr {
 				panic(fmt.Errorf("method %q.%q: paged return type %q.Items is not an array of pointers", m.Service, m.Name(), listType.Name()))
 			}
+
+			if itemsType.Kind() != reflect.Slice || itemsType.Elem().Kind() != reflect.Pointer {
```
aojea, line 238: the previous check and this are confusing, first the names Ptr vs Pointer, it seems Ptr was deprecated and now means the same as Pointer https://pkg.go.dev/reflect#pkg-constants and https://github.com/golang/go/issues/47651

if Ptr and Pointer is the same this is just the previous check with an OR instead of an AND , so this is more restrictive than the previous that requires both to be different and now, we panic if at least one is different

Should you remove the previous check then?

bowei: ok, looks like the real bug was a conditional that was wrong... fixing

PR #143, opened 2023-10-20
Fixes support for TcpRoute in the AppNet APIs

bowei:
Fix various bugs needed to make TcpRoute resource work
    
    - Operations needs to be implemented
    - Various mismatches in the interface with the compute API

### DIFF
--- a/e2e/tcproute_test.go
+++ b/e2e/tcproute_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2023 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"github.com/kr/pretty"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/networkservices/v1"
+)
+
+func TestTcpRoute(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	bs := &compute.BackendService{
+		Name:                resourceName("bs1"),
+		Backends:            []*compute.Backend{},
+		LoadBalancingScheme: "INTERNAL_SELF_MANAGED",
+	}
+	bsKey := meta.GlobalKey(bs.Name)
+
+	t.Cleanup(func() {
+		err := theCloud.BackendServices().Delete(ctx, bsKey)
+		t.Logf("bs delete: %v", err)
+	})
+
+	// TcpRoute needs a BackendService to point to.
+	err := theCloud.BackendServices().Insert(ctx, bsKey, bs)
+	t.Logf("bs insert: %v", err)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Current API does not support the new URL scheme.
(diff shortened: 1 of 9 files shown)


e2e/tcproute_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"github.com/kr/pretty"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/networkservices/v1"
+)
+
+func TestTcpRoute(t *testing.T) {
```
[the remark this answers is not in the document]
bowei: These are run manually for now. We have a background investigation to setup CI, but it was a bit complex.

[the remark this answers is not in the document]
bowei: (e.g. it is work in progress)

bowei:
This should be ready for review

e2e/tcproute_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Insert
+	err = theCloud.TcpRoutes().Insert(ctx, key, tcpr)
+	t.Logf("tcproutes insert: %v", err)
+	if err != nil {
+		t.Fatalf("Insert() = %v", err)
+	}
+
+	// Get
+	tcpRoute, err := theCloud.TcpRoutes().Get(ctx, key)
```
aojea, line 91: don't you need to assert that what you get is really what you inserted?

pkg/cloud/meta/service.go
```diff
@@ -116,6 +117,19 @@ func (i *ServiceInfo) ListItemName() string {
 	return "Items"
 }
 
+func (i *ServiceInfo) NetworkServicesFmt() string {
```
aojea, line 120: Why don't you use the Stringer https://go.dev/tour/methods/17 implementing the `String()`method for this?

pkg/cloud/op_networkservices.go
```diff
@@ -0,0 +1,84 @@
...
+
+type networkServicesOperation struct {
+	s         *Service
+	projectID string
+	key       *meta.Key
+	err       error
+}
+
+func (o *networkServicesOperation) String() string {
+	return fmt.Sprintf("networkServicesOperation{%q, %v}", o.projectID, o.key)
```
aojea, line 39: `o.key` or `o.key.String()`

pkg/cloud/op_networkservices.go
```diff
@@ -0,0 +1,84 @@
...
+	return &RateLimitKey{
+		ProjectID: o.projectID,
+		Operation: "Get",
+		Service:   "Operations",
+		Version:   meta.VersionGA,
+	}
+}
+
+func (o *networkServicesOperation) error() error {
+	return o.err
```
aojea, line 83: do we care about thread safety here?

pkg/cloud/service.go
```diff
@@ -41,32 +47,122 @@ type Service struct {
...
 
+type networkServiceOpURLParseResult struct {
+	projectID string
+	key       *meta.Key
+}
+
+// parseNetworkServiceOpURL parses the URL of the network services operation.
+// This is different than the `compute` API paths.
+func parseNetworkServiceOpURL(name string) (*networkServiceOpURLParseResult, error) {
+	// Format: projects/<projectID>/locations/global/operations/<Name>
```
aojea, line 154: do we need to do a sanity check that split[0] == projects, split[2] == locations , .... ?

e2e/tcproute_test.go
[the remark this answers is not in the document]
aojea: these tests requires to perform GCE calls or are mocked?

aojea:
mostly questions for self education

pkg/cloud/service.go
bowei: adding.

pkg/cloud/op_networkservices.go
bowei: library is mostly thread agnostic -- you are expected to guard for threads outside this code.

pkg/cloud/op_networkservices.go
bowei: changed to use %s, which calls String()

pkg/cloud/meta/service.go
bowei: Don't completely follow your comment, can you explain more?

PR #147, opened 2023-11-02
add error test cases for transitive closure

kl52752:


pkg/cloud/rgraph/algo/trclosure/transitive_closure_test.go
```diff
@@ -237,7 +239,76 @@ func TestTransitiveClosure(t *testing.T) {
 			},
 			want: wantGraph{ids: []string{"a", "b", "c"}},
 		},
-		// TODO(bowei): add error test cases.
+		{
+			name:  "don't traverse error nodes",
```
bowei, line 243: name: "nodes with error results in err"

pkg/cloud/rgraph/algo/trclosure/transitive_closure_test.go
```diff
@@ -237,7 +239,76 @@ func TestTransitiveClosure(t *testing.T) {
...
+				addNode("b", []string{"d", "e"}, nodeIsInError)
+				addNode("d", []string{"a"})
+				addNode("e", []string{"b"})
+				return g
+			},
+			want:    wantGraph{ids: []string{"a", "b", "c"}},
+			wantErr: true,
+		},
+		{
+			name:  "don't traverse error nodes - child in error",
```
bowei, line 257: I think we need to define the behavior in the case where there is an error encountered. Do we expect that the traverse will continue and we get as much of the graph as we can OR will the traversal be canceled and the result be just whatever was fetched up around the time we detected the error and tried to stop?

If we can add this to the doc strings as well and also change the documented name of the test cases, this is will be more obvious to maintainers.

kl52752: yes, you are right there is no point checking the output on the error.

pkg/cloud/rgraph/algo/trclosure/transitive_closure_test.go
kl52752: done

kl52752:
/cc @mag-kol
//cc @tangenti

pkg/cloud/rgraph/algo/trclosure/transitive_closure_test.go
```diff
@@ -237,7 +239,59 @@ func TestTransitiveClosure(t *testing.T) {
...
 		},
-		// TODO(bowei): add error test cases.
+		{
+			name:  "nodes with error results in err",
+			cloud: mockCloud,
+			graph: func() *rgraph.Builder {
+				g := rgraph.NewBuilder()
+				addNodeToGraph(g, "a", []string{"b", "c"})
+				addNode("b", []string{"d", "e"}, nodeIsInError)
+				addNode("d", []string{"a"})
```
bowei, line 249: Do we need these additional nodes if we are testing the error cases? We should make the error cases simple so people who maintain this later don't have to figure out whether or not the extra setup was part of the test? This comment applies to all of the test cases being added.

pkg/cloud/rgraph/rnode/fake/builder.go
```diff
@@ -60,7 +62,7 @@ func (b *Builder) SetResource(u rnode.UntypedResource) error {
 func (b *Builder) SyncFromCloud(ctx context.Context, gcp cloud.Cloud) error {
 	// TODO(bowei): add ability to inject errors for the SyncFromCloud.
```
bowei, line 63: Please fix this comment if it is no longer relevant.

pkg/cloud/rgraph/algo/trclosure/transitive_closure_test.go
```diff
@@ -237,7 +239,40 @@ func TestTransitiveClosure(t *testing.T) {
...
 			want: wantGraph{ids: []string{"a", "b", "c"}},
 		},
-		// TODO(bowei): add error test cases.
+		{
+			name:  "nodes with error results in err",
+			cloud: mockCloud,
+			graph: func() *rgraph.Builder {
+				g := rgraph.NewBuilder()
+				addNodeToGraph(g, "a", []string{"b", "c"})
+				addNode("b", []string{"d", "e"}, nodeIsInError)
```
bowei, line 248: we need to fix these references to non-existant nodes? the same comment applies to all of the new test cases

kl52752: done

PR #155, opened 2023-11-21
Fix mock hooks, add options argument, add compile-time verification

panslava:
- After changes in mocks interfaces, hooks were not updated with new options argument. That resulted in compile error, for users of custom hooks
- Add compile time verification that hook implements interface, for every custom hook implementation.

/assign @bowei @spencerhance


panslava:
Fun (sad) fact, I almost manually editted all this file, function names were not consistent enough to write some complicate regex......

PR #156, opened 2023-11-21
Add support for all array elements

kl52752:
When defining type traits we needed a way to match all arrays indexes to access arrays struct's fields.

As an example why this change is needed and how it will be used I added Backend Service support where Backends, an array which is required, has some optional fields.


pkg/cloud/api/path.go
```diff
@@ -29,20 +29,28 @@ import (
...
 )
 
 // Field returns the path extended with a struct field reference.
 func (p Path) Field(name string) Path {
 	return append(p, string(pathField)+name)
 }
 
+// Field returns the path extended with a all array element symbol.
+// This is needed for matching array's nested fields.
+func (p Path) AllElem() Path {
```
bowei, line 50: ArrayElements

bowei:
split off the change for backendservice into a separate commit

kl52752:
> split off the change for backendservice into a separate commit

done, I will post PR with BS changes after this one is closed.

pkg/cloud/api/path.go
```diff
@@ -92,6 +123,23 @@ func (p Path) HasPrefix(prefix Path) bool {
...
+	if len(path) != 2 {
+		return false
+	}
+	if path[0] != pathSliceIndex {
+		return false
+	}
+	return path[1] == arrayAllElem
+}
+
+func isArrayIndex(path string) bool {
```
bowei, line 136: We should always keep the same naming convention as the constant to make it easier to understand.

change to `func isPathSliceIndex(element string)`

The string argument is not a "Path" type.

pkg/cloud/api/path.go
```diff
@@ -71,6 +79,26 @@ func (p Path) Equal(other Path) bool {
 	return true
 }
 
+// Similar returns true if other is the same path.
+// For arrays this function does not check indices.
+func (p Path) Similar(other Path) bool {
```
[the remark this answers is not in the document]
bowei: How is this method used?

pkg/cloud/api/path.go
```diff
@@ -29,20 +29,28 @@ import (
...
 // - "*" is a pointer deref.
+// - "#" is all array elements reference
 type Path []string
 
 const (
 	pathField      = '.'
 	pathSliceIndex = '!'
 	pathMapIndex   = ':'
 	pathPointer    = '*'
+	arrayAllElem   = '#'
```
bowei, line 40: anyIndex

pkg/cloud/api/path.go
```diff
@@ -92,6 +123,23 @@ func (p Path) HasPrefix(prefix Path) bool {
 	return true
 }
 
+func isAllElemPrefix(path string) bool {
+	if len(path) != 2 {
```
bowei, line 127: func isAnyIndexElement(element string) bool { return element == string(pathSliceIndex) + string(anyIndex)

pkg/cloud/api/path.go
```diff
@@ -82,6 +110,9 @@ func (p Path) HasPrefix(prefix Path) bool {
 
 	var i int
 	for i = range prefix {
+		if isAllElemPrefix(prefix[i]) && isArrayIndex(p[i]) {
```
bowei, line 113: Let's add a small comment

// anyIndex matches any index.

pkg/cloud/api/path.go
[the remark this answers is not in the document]
kl52752: I think this function should be used in `fieldTrait` function.
This function checks is search path is in fieldTraits array where we define fields types. The check is done by HasPrefix and IMO it's not good because for structure types we cannot define custom "rules" per fields if we defined rule for structure itself. 
The example :
We have structs:
* TCPRout {
bs BackendService // optional
} 
* BackendService {
name Name // required
rate Rate //optional
}  
We have a TCP route struct which uses BackendService struct as an optional field. BackendService has some optional and required fields like Name. 
In current solution if we are using HAsPrefix to check 2 paths and we define that BackendService is optional field then for path tcproute.backendService.Name we might match the parent tcproute.backendService which is marked as optional field not the field itself. 

And in this function `Equal` will not work because of the exact matching. I. not sure if we need Equal and Similar or just Similar (Equal with logic change to support AnyArrayIndex)  but for sure we need to match paths for arrays :)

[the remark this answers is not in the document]
bowei: Can you give a couple of examples with all of the values filled in (maybe as some unit test cases). I'm not 100% following your example (this is probably an issue w/ my understanding).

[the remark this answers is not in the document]
bowei: Is this resolved by sorting the trait matching to prefer matching the longest prefix?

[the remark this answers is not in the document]
kl52752: > Is this resolved by sorting the trait matching to prefer matching the longest prefix?

yes this could be but right now by default we have required fields and we set OutputOnly and AllowZeroValue for some of them. With this approach we would have to define traits for all fields and current implementation with string comparison might be slow (there is a TODO to change this to tree)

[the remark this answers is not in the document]
kl52752: But also for traits we don't want prefix match but path match right?

pkg/cloud/api/path_test.go
```diff
@@ -59,6 +59,8 @@ func TestPathEqual(t *testing.T) {
 		{a: Path{}.Pointer(), b: Path{}.Pointer(), want: true},
 		{a: Path{}, b: Path{}.Pointer(), want: false},
 		{a: Path{}.Index(0), b: Path{}.MapIndex(0), want: false},
+		{a: Path{}.Index(0), b: Path{}.AnyIndex(), want: false},
+		{a: Path{}.MapIndex(0), b: Path{}.AnyIndex(), want: false},
```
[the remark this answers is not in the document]
bowei: Oh I see -- we want equal to be literally equal and add a "match" operator.

pkg/cloud/api/path.go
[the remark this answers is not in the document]
bowei: I think I understand now that you want to keep .Equal to be absolute == and have a separate operation that is interpret match.

pkg/cloud/api/path.go
```diff
@@ -71,6 +78,26 @@ func (p Path) Equal(other Path) bool {
 	return true
 }
 
+// Similar returns true if other is the same path.
+// For arrays this function does not check indices.
+func (p Path) Similar(other Path) bool {
```
bowei, line 83: I would prefer to call this Match(). Similar() in my mind is a bit ambiguous what it means.

pkg/cloud/api/path.go
```diff
@@ -71,6 +78,26 @@ func (p Path) Equal(other Path) bool {
 	return true
 }
 
+// Similar returns true if other is the same path.
```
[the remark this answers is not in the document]
bowei: Let's do this for the docstring

```
// Match the given path, interpreting wildcard matching for the comparison.
// Use Equal() to compare Paths for absolute equality (no wildcard interpretation).
```

pkg/cloud/api/path.go
```diff
@@ -71,6 +78,26 @@ func (p Path) Equal(other Path) bool {
...
+// For arrays this function does not check indices.
+func (p Path) Similar(other Path) bool {
+	if len(p) != len(other) {
+		return false
+	}
+	for i := range p {
+		if isAnyIndexElement(p[i]) && isArrayIndex(other[i]) {
```
bowei, line 88: add 

// Wildcard match

pkg/cloud/api/path.go
```diff
@@ -92,6 +123,15 @@ func (p Path) HasPrefix(prefix Path) bool {
 	return true
 }
 
+// isAnyIndexElement checks if path matches slice path with any index
```
bowei, line 126: // isAnyIndexElement is true if this is an AnyIndex element.

pkg/cloud/api/path.go
```diff
@@ -92,6 +123,15 @@ func (p Path) HasPrefix(prefix Path) bool {
...
+func isAnyIndexElement(element string) bool {
+	return element == string(pathSliceIndex)+string(anyIndex)
+}
+
+func isArrayIndex(path string) bool {
```
bowei, line 131: * Change name to `isSliceIndex`. You should always match the existing naming convention (e.g Slice vs Array) to make it easier for people to understand the code as a maintainer.
* Change parameter name to `element`. You are not getting a `path` passed into the func.

Change docstring comment: `// isArrayIndex returns true element is a SliceIndex`

bowei:
Make sure you update your commit comment to match the updated code

kl52752:
> Make sure you update your commit comment to match the updated code

done

kl52752:
@bowei I added support for Map indices. At first I thought that I will do this in new PR but since you mentioned it in some comment I added functions + tests for that. PTAL

pkg/cloud/api/path.go
```diff
@@ -29,20 +29,32 @@ import (
 // - "!" is a slice index
 // - ":" is a map key
 // - "*" is a pointer deref.
+// - "#" is all array elements reference
```
bowei, line 32: I think we should have this in a different comment block as this character is not used as a prefix:

```
const (
  ...
  pathPointer = '#'

  // anyIndex is a wildcard index used instead of the number (slice) or string key (map).
  anyIndex = '#'
)
```

pkg/cloud/api/path.go
```diff
@@ -58,7 +70,11 @@ func (p Path) Pointer() Path {
 	return append(p, string(pathPointer))
 }
 
-// Equal returns true if other is the same path.
+// Equal returns true if other is the same path. Note that this equality
+// comparison does NOT interpret wildcard matches, e.g. .AnyIndex is only
+// Equal() to .AnyIndex, not Equal to .Index(x).
+//
+// Use Match() to match interpreting wildcards.
```
bowei, line 77: Make this more clear to the user:

Use Match() **instead** to match

pkg/cloud/api/path.go
```diff
@@ -71,6 +87,33 @@ func (p Path) Equal(other Path) bool {
 	return true
 }
 
+// Match the given path, interpreting wildcard matching for the comparison. Use
+// Equal() to compare Paths for absolute equality (no wildcard interpretation).
```
bowei, line 91: Use Equal() **instead** to compare...

pkg/cloud/api/path.go
```diff
@@ -71,6 +87,33 @@ func (p Path) Equal(other Path) bool {
...
+		}
+		if p[i] != other[i] {
+			return false
+		}
+	}
+	return true
+}
+
 // HasPrefix returns true if prefix is the prefix of this path.
 func (p Path) HasPrefix(prefix Path) bool {
```
bowei, line 118: Let's add to the docstring

// HasPrefix returns true if prefix matches this path. Prefix
// uses Match() semantics for wildcards.

pkg/cloud/api/path.go
bowei: Now that I think about it, isn't it much simpler to do this:

```
anySliceIndex = string(pathSliceIndex) + "#"
anyMapIndex = string(pathMapIndex) + "#"
```

then we don't need all the extra code below?

pkg/cloud/api/path.go
bowei: Let's do this:

```
// isMatch returns true if a matches b, interpreting wildcards. This func is symmetric.
func isMatch(a, b string) bool {
  if a == anySliceIndex && isSliceIndex(b) || 
     a == anyMapIndex && isMapIndex(b) || 
     b == anySliceIndex && isSliceIndex(a) || 
     b == anyMapIndex && isMapIndex(a) {
    return true
  }
  return a == b
}
```

pkg/cloud/api/path.go
```diff
@@ -82,6 +125,13 @@ func (p Path) HasPrefix(prefix Path) bool {
 
 	var i int
 	for i = range prefix {
+		// anyIndex matches any index.
```
bowei, line 128: Use isMatch() as mentioned above.

pkg/cloud/api/path.go
```diff
@@ -29,20 +29,32 @@ import (
...
 )
 
 // Field returns the path extended with a struct field reference.
 func (p Path) Field(name string) Path {
 	return append(p, string(pathField)+name)
 }
 
+// AnySliceIndex returns a path extended to match any slice index.
+func (p Path) AnySliceIndex() Path {
+	return append(p, string(pathSliceIndex)+string(anyIndex))
```
bowei, line 50: return append(p, anySliceIndex)

pkg/cloud/api/path.go
```diff
@@ -29,20 +29,32 @@ import (
...
+}
+
+// AnyMapIndex returns a path extended to match any map index.
+func (p Path) AnyMapIndex() Path {
+	return append(p, string(pathMapIndex)+string(anyIndex))
```
bowei, line 55: return append(p, anyMapIndex)

pkg/cloud/api/path.go
kl52752: is match is bidirectional if we have 2 paths one or the other might be "anyIndex". In HasPrefix only prefix can be anyIndex and we expect `p` to be slice/map index with actual value.
I will add check for values to `isSliceIndex` and `isMapIndex` funcions.
WDYT?

pkg/cloud/api/path.go
```diff
@@ -92,6 +147,20 @@ func (p Path) HasPrefix(prefix Path) bool {
 	return true
 }
 
+// isSliceIndex returns true if the element is a SliceIndex
+func isSliceIndex(element string) bool {
+	if len(element) < 3 && element[0] != pathSliceIndex {
```
bowei, line 152: why 3? isn't !1 a valid index?

pkg/cloud/api/path.go
```diff
@@ -71,7 +91,36 @@ func (p Path) Equal(other Path) bool {
...
+		}
+		if p[i] != other[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// HasPrefix returns true if prefix is the prefix of this path. Prefix uses
+// Match() semantics for wildcards.
```
bowei, line 123: Let make this predicate symmetric and add (`... wildcards. This predicate is symmetric.`) unless we have a good reason to make it non-symmetric, e.g. there is a use case. This is easier to use and understand and is implied by the statement, this uses Match() semantics, as Match is symmetric.

pkg/cloud/api/path.go
```diff
@@ -92,6 +147,20 @@ func (p Path) HasPrefix(prefix Path) bool {
 	return true
 }
 
+// isSliceIndex returns true if the element is a SliceIndex
+func isSliceIndex(element string) bool {
```
bowei, line 151: Add unit tests to cover these new functions.

What is the coverage for the newly added code.

pkg/cloud/api/path_test.go
```diff
@@ -67,6 +72,39 @@ func TestPathEqual(t *testing.T) {
 	}
 }
 
+func TestPathSimilar(t *testing.T) {
```
bowei, line 75: Need to change the test names to match the new names of the funcs..

pkg/cloud/api/path_test.go
```diff
@@ -67,6 +72,39 @@ func TestPathEqual(t *testing.T) {
...
+		{a: Path{}.AnySliceIndex(), b: Path{}.Index(8), want: true},
+		{a: Path{}.AnySliceIndex().Field("abcd").Field("cde"), b: Path{}.Index(2).Field("abcd"), want: false},
+		{a: Path{}.AnySliceIndex(), b: Path{}.MapIndex(0), want: false},
+		{a: Path{}.AnyMapIndex(), b: Path{}.MapIndex(1), want: true},
+		{a: Path{}.AnyMapIndex(), b: Path{}.Index(9), want: false},
+		{a: Path{}.AnyMapIndex(), b: Path{}.AnySliceIndex(), want: false},
+	} {
+		got := tc.a.Match(tc.b)
+		if got != tc.want {
+			t.Errorf("Equal(%s, %s) = %t, want %t", tc.a, tc.b, got, tc.want)
```
bowei, line 103: Match(

pkg/cloud/api/path.go
kl52752: fixed, sorry for bug, added unit tests

pkg/cloud/api/path.go
kl52752: I don't really think that symmetric is a good word here. 
For me symmetric means that p.HasPrefix(prefix) == prefix.HasPrefix(p) but this condition will not be true because 
in `HasPrefix` we expect that path length >= prefix length and that means that this function is not symmetric. 
By symmetric you mean that both prefix and that can have wildcards?
IMO prefix should be generic (and can use wildcard because we don't care about real indexes) and path should be a regular path extracted from the object so it has to have real indexes and map's key. 

We cannot use path.Match here because Match check path lengths.
I extracted isMatch function which compares the elements (and should be used in for loop) in case we wanted to use it here.

pkg/cloud/api/path.go
kl52752: ok      github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/api (cached)        coverage: 85.9% of statements

pkg/cloud/api/path.go
kl52752: done

pkg/cloud/api/path.go
bowei: make sense -- this is not symmetric

PR #158, opened 2023-11-22
add PlaceholderType

kl52752:
PlaceholderType is used to represent GCE resource type versions that either
don't exist (e.g. there is not alpha version of the given resource) or we do
not intend to use (e.g. we omitted the type/version from being included in
pkg/cloud/meta).

### DIFF
--- a/pkg/cloud/api/check_test.go
+++ b/pkg/cloud/api/check_test.go
@@ -258,6 +258,7 @@ func TestCheckSchema(t *testing.T) {
 		wantErr bool
 	}{
 		{name: "ok", t: reflect.TypeOf(&okSt{})},
+		{name: "PlaceholderType is ok", t: reflect.TypeOf(&PlaceholderType{})},
 		{name: "fails cycle check", t: reflect.TypeOf(&rec2{}), wantErr: true},
 		{name: "fails type check", t: reflect.TypeOf(&badSt{}), wantErr: true},
 		{name: "fails type check bad fields", t: reflect.TypeOf(&badStFieldsBad{}), wantErr: true},
--- a/pkg/cloud/api/mutable_resource.go
+++ b/pkg/cloud/api/mutable_resource.go
@@ -54,6 +54,21 @@ func (e *ConversionError) Error() string {
 	return fmt.Sprintf("ConversionError: missing fields %v", e.MissingFields)
 }
 
+// useOfPlaceholderTypeError is raised when code attempts to convert or operate
+// on a Resource type that is a placeholder. For example, given:
+//
+//	 Resource[ga.Res, /*alpha*/ PlaceholderType, beta.Res]
+//
+//	if the code tries to convert the Resource to the Alpha type,
+//	the operation will fail with this error.
+type useOfPlaceholderTypeError struct {
+	msg string
+}
+
+func (m useOfPlaceholderTypeError) Error() string {
+	return "UseOfPlaceholderTypeError: " + m.msg
+}
+
 // MissingField describes a field that was lost when converting between API
 // versions due to the field not being present in struct.
 type MissingField struct {
@@ -168,6 +183,10 @@ type mutableResource[GA any, Alpha any, Beta any] struct {
 }
 
 func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
+	if isPlaceholderType(u.ga) {
+		return fmt.Errorf("GA has unsupported type (type is %T)", u)
+	}
+
 	err := checkSchema(reflect.TypeOf(&u.ga))
 	if err != nil {
 		return err
@@ -180,6 +199,7 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
 	if err != nil {
 		return err
 	}
+	// TODOD(kl52752) Add validation that GA is a subset of Beta and Alpha.
 	return nil
 }
 
@@ -202,40 +222,52 @@ func (u *mutableResource[GA, Alpha, Beta]) postAccess(srcVer meta.Version, flags
 	switch srcVer {
 	case meta.VersionGA:
 		src = reflect.ValueOf(&u.ga)
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.alpha),
-			copyHelper: func() error { return u.typeTrait.CopyHelperGAtoAlpha(&u.alpha, &u.ga) },
-			errors:     &u.errors[GAToAlphaConversion],
-		})
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.beta),
-			copyHelper: func() error { return u.typeTrait.CopyHelperGAtoBeta(&u.beta, &u.ga) },
-			errors:     &u.errors[GAToBetaConversion],
-		})
+		if !isPlaceholderType(u.alpha) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.alpha),
+				copyHelper: func() error { return u.typeTrait.CopyHelperGAtoAlpha(&u.alpha, &u.ga) },
+				errors:     &u.errors[GAToAlphaConversion],
+			})
+		}
+		if !isPlaceholderType(u.beta) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.beta),
+				copyHelper: func() error { return u.typeTrait.CopyHelperGAtoBeta(&u.beta, &u.ga) },
+				errors:     &u.errors[GAToBetaConversion],
+			})
+		}
 	case meta.VersionAlpha:
 		src = reflect.ValueOf(&u.alpha)
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.ga),
-			copyHelper: func() error { return u.typeTrait.CopyHelperAlphaToGA(&u.ga, &u.alpha) },
-			errors:     &u.errors[AlphaToGAConversion],
-		})
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.beta),
-			copyHelper: func() error { return u.typeTrait.CopyHelperAlphaToBeta(&u.beta, &u.alpha) },
-			errors:     &u.errors[AlphaToBetaConversion],
-		})
+		if !isPlaceholderType(u.ga) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.ga),
+				copyHelper: func() error { return u.typeTrait.CopyHelperAlphaToGA(&u.ga, &u.alpha) },
+				errors:     &u.errors[AlphaToGAConversion],
+			})
+		}
+		if !isPlaceholderType(u.beta) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.beta),
+				copyHelper: func() error { return u.typeTrait.CopyHelperAlphaToBeta(&u.beta, &u.alpha) },
+				errors:     &u.errors[AlphaToBetaConversion],
+			})
+		}
 	case meta.VersionBeta:
 		src = reflect.ValueOf(&u.beta)
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.ga),
-			copyHelper: func() error { return u.typeTrait.CopyHelperBetaToGA(&u.ga, &u.beta) },
-			errors:     &u.errors[BetaToGAConversion],
-		})
-		conversions = append(conversions, convert{
-			dest:       reflect.ValueOf(&u.alpha),
-			copyHelper: func() error { return u.typeTrait.CopyHelperBetaToAlpha(&u.alpha, &u.beta) },
-			errors:     &u.errors[BetaToAlphaConversion],
-		})
+		if !isPlaceholderType(u.ga) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.ga),
+				copyHelper: func() error { return u.typeTrait.CopyHelperBetaToGA(&u.ga, &u.beta) },
+				errors:     &u.errors[BetaToGAConversion],
+			})
+		}
+		if !isPlaceholderType(u.alpha) {
+			conversions = append(conversions, convert{
+				dest:       reflect.ValueOf(&u.alpha),
+				copyHelper: func() error { return u.typeTrait.CopyHelperBetaToAlpha(&u.alpha, &u.beta) },
+				errors:     &u.errors[BetaToAlphaConversion],
+			})
+		}
 	}
 
 	if flags&postAccessSkipValidation == 0 {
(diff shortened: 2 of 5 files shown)


pkg/cloud/api/type.go
```diff
@@ -20,6 +20,15 @@ import (
 	"reflect"
 )
 
+// UntypedResourceType should be used to represent missing GCE resources
+// in type traits functions.
+// Every GCE resource has Name and SelfLink so check schema for this
+// object will pass.
```
bowei, line 26: give a usage example. 

Use 'fill paragraph' to format the comments (fill to 78 columns). Note: this is M-q in Emacs, 'gqap' in Vim.

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,13 @@ func (e *ConversionError) Error() string {
 	return fmt.Sprintf("ConversionError: missing fields %v", e.MissingFields)
 }
 
+// MissingTypeError is returned when type traits has missing type.
+type MissingTypeError struct{}
```
bowei, line 58: We should make it easier to understand when this error arises:

```
// UseOfPlaceholderTypeError is raised when code attempts to convert or operate
// on a Resource type that is a placeholder. For example, given:
//
//   Resource[ga.Res, /*alpha*/ PlaceholderType, beta.Res]
// 
//  if the code tries to convert the Resource to the Alpha type,
//  the operation will fail with this error.
```

pkg/cloud/api/check_test.go
```diff
@@ -258,6 +258,7 @@ func TestCheckSchema(t *testing.T) {
 		wantErr bool
 	}{
 		{name: "ok", t: reflect.TypeOf(&okSt{})},
+		{name: "ok-untyped", t: reflect.TypeOf(&MissingResourceType{})},
```
bowei, line 261: Change to `{name: "PlaceholderType is ok"}`

You should always align the names so people have an easy time following the code:

Here we has "untyped", then the type is "MissingResourceType". I put below that we
should change the name to PlaceholderType. Let's make sure all of the references and
places where we use things use the same kinds of names.

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,13 @@ func (e *ConversionError) Error() string {
...
+
+func (m *MissingTypeError) Error() string {
```
bowei, line 60: It would be better to be able to put more details in the error message to make it easy for someone who hits this error to know immediately where the error is coming from.

Are we going to be checking specifically for MissingTypeError? If not, can we just use fmt.Errorf() and give the user more meaningful errors than just MissingTypeError?

pkg/cloud/api/type_test.go
```diff
@@ -80,3 +80,60 @@ func TestTypeIsShortcuts(t *testing.T) {
 		}
 	}
 }
+
+func TestMissingType(t *testing.T) {
```
bowei, line 84: TestPlaceholderType

pkg/cloud/api/mutable_resource.go
```diff
@@ -272,24 +291,51 @@ func (u *mutableResource[GA, Alpha, Beta]) AccessBeta(f func(x *Beta)) error {
...
+	case notConvertibleTo(alphaErr, u.alpha) && completeType(betaErr, u.beta):
 		return meta.VersionBeta, nil
 	default:
 		return meta.VersionGA, fmt.Errorf("indeterminant version (ga=%v, alpha=%v, beta=%v)", gaErr, alphaErr, betaErr)
 	}
 }
 
 func (u *mutableResource[GA, Alpha, Beta]) ToGA() (*GA, error) {
+	if IsTypeMissing(u.ga) {
+		return nil, &MissingTypeError{}
```
bowei, line 337: See comment above -- we should return a better error, otherwise people will end up with a not-so- useful error message in their logs.

pkg/cloud/api/mutable_resource.go
```diff
@@ -272,24 +291,51 @@ func (u *mutableResource[GA, Alpha, Beta]) AccessBeta(f func(x *Beta)) error {
 	return u.postAccess(meta.VersionBeta, 0)
 }
 
+// completeType means that the object does not have conversion errors and its
+// type is not missing.
+func completeType(err error, obj interface{}) bool {
```
bowei, line 296: See comment below, I think there is a simpler way to do this.

pkg/cloud/api/mutable_resource.go
```diff
@@ -272,24 +291,51 @@ func (u *mutableResource[GA, Alpha, Beta]) AccessBeta(f func(x *Beta)) error {
...
+func convertibleTo(err error, obj interface{}) bool {
+	return err == nil || IsTypeMissing(obj)
+}
+
+// convertibleTo means that the object has conversion errors or its type is
+// missing.
+func notConvertibleTo(err error, obj interface{}) bool {
+	return err != nil || IsTypeMissing(obj)
+}
+
```
bowei, line 311: We should document the version conversion logic now that it is more complex due to PlaceholderType

```
// ImpliedVersion returns the implied version of the underlying resource.
// This is determined by the convertibility of the resource.
//
//  y = Can convert to this version.
//  n = Can't convert (note: nothing can convert to PlaceholderType)
//
//  | GA| A | B |
//  -------------
//  | y | y | y | GA
//  | y | y | n | Alpha
//  | y | n | y | Beta
//  | y | n | n | GA (Note: this is somewhat weird as GA /should/ convert)
//  | n | y | y | Error (can't determine version)
//  | n | y | n | Alpha
//  | n | n | y | Beta
//  | n | n | n | Error (can't determine version)
```

pkg/cloud/api/mutable_resource.go
```diff
@@ -272,24 +291,51 @@ func (u *mutableResource[GA, Alpha, Beta]) AccessBeta(f func(x *Beta)) error {
...
 func (u *mutableResource[GA, Alpha, Beta]) ImpliedVersion() (meta.Version, error) {
 	_, gaErr := u.ToGA()
 	_, alphaErr := u.ToAlpha()
 	_, betaErr := u.ToBeta()
 
-	switch {
-	case gaErr == nil && alphaErr == nil && betaErr == nil:
+	// If type is GA and converts to Alpha and Beta without error return GA.
```
bowei, line 317: It seems like the code you have here is redundant (I might be missing something)

`ToGA`, `ToAlpha`, `ToBeta` already return an error when the type that is being converted to is `PlaceholderType`. This means that we don't have to have an additional level of checking to implement the logic table above.

To make the code match the logic table almost exactly, I recommend something like this:

```
	type v struct{ ga, alpha, beta bool }
	result := v{ga: gaErr != nil, alpha: alphaErr != nil, beta: betaErr != nil}
	switch result {
	case v{true, true, true}:
    return meta.VersionGA, nil
	case v{true, true, false}:
    return meta.VersionAlpha, nil
		//...
	}
```

kl52752: I added those functions for better understanding why we imply such version like `GA needs to be complete type and should convert to Alpha and Beta`. where complete type means that GA type cannot be `PlaceholderType` and convert to means that there is no error from conversion or type is `PlaceholderType`. It is more readable for me :) thank `err != nil` 

In your proposed solution we are missing the information why conversion failed.
For GA we should return GA when Alpha and Beta convert without errors or are `PlaceholderType` in other case where conversion fails due to some other reason we should return error.

pkg/cloud/api/mutable_resource.go
kl52752: //  | y | n | n | GA (Note: this is somewhat weird as GA /should/ convert)
-- I think we should clearly said that the only case where we will imply GA version and conversion Alpha and Beta fails is where those types are `PlaceholderType` in other cases we should return error.
Also according to previous code we will deduce Alpha version only when GA and Beta conversion fails so
`| y | y | n | Alpha` and `| y | n | y | Beta` are not true

Proposition 
//  y = Can convert to this version. 
//  n = Can't convert (note: conversion of PlaceholderType will return 
// UseOfPlaceholderTypeError. This is a special case not covered here)
//  y/p = can convert or conversion returned UseOfPlaceholderTypeError
//  p = conversion returned UseOfPlaceholderTypeError

//  | GA| A | B |
//  -------------
//  | y | y/p | y/p | GA
//  | n | y   | n   | Alpha
//  | n | n   | y   | Beta
//  | y | n   | n   | Error (can't determine version)
//  | n | n   | n   | Error (can't determine version)
//  | p | y/n | y/n | Error (GA cannot be PlaceholderType)

pkg/cloud/api/mutable_resource.go
bowei: I find that the tabular format is easier to understand. 

In your code below, I have to understand both what convertibleTo does, the first argument is a generic error type (that could be anything) and the second is a type that is passed to a typecheck. It's non-trivial to check that the behavior given above is matching the code (you can do it by inspection) rather than having to reason about the `if` and `switch` statements below.

kl52752: I don't mind adding tabular format (when we agree on the result) I want also to add functions that are descriptive.

bowei:
Fix the title of this PR and the commit message to match the new code and renamed types, etc.

pkg/cloud/api/mutable_resource.go
kl52752: the discussion was moved offline. Resolving this thread

pkg/cloud/api/mutable_resource.go
kl52752: The discussion was moved offline. Resolving this thread

pkg/cloud/api/mutable_resource.go
kl52752: changed

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,19 @@ func (e *ConversionError) Error() string {
...
 }
 
+// UseOfPlaceholderTypeError is raised when code attempts to convert or operate
+// on a Resource type that is a placeholder. For example, given:
+//
+//	 Resource[ga.Res, /*alpha*/ PlaceholderType, beta.Res]
+//
+//	if the code tries to convert the Resource to the Alpha type,
+//	the operation will fail with this error.
+type UseOfPlaceholderTypeError struct{}
```
bowei, line 64: I would prefer:

type UseOfPlaceholderTypeError struct {
  msg string
}

func (e UseOfPlaceholderTypeError) Error() string {
  return "UseOfPlaceholderTypeError: " + e.msg
}

pkg/cloud/api/mutable_resource.go
```diff
@@ -272,21 +302,38 @@ func (u *mutableResource[GA, Alpha, Beta]) AccessBeta(f func(x *Beta)) error {
...
+// -----------------
+// | y | y | y | GA
+// | n | y | y | Beta
+// | n | y | n | Beta
+// | n | n | y | Alpha
+// | n | n | n | Error
 func (u *mutableResource[GA, Alpha, Beta]) ImpliedVersion() (meta.Version, error) {
 	_, gaErr := u.ToGA()
-	_, alphaErr := u.ToAlpha()
+	if gaErr == nil {
```
bowei, line 323: This code looks to be wrong?

bowei:
See comment -- you need to fix the commit message and the PR title.

pkg/cloud/api/mutable_resource.go
kl52752: We are checking the conversion GA -> Beta and Alpha in CheckSum so I thought that there is no need to do this here because we enforce this is UT.
So I only check toGA and if this succeed I don't check toBeta and toAlpha because they can fail due to PlaceholderType and in CheckSum I checked that GA is a subset of B and A.

pkg/cloud/api/mutable_resource.go
kl52752: done

pkg/cloud/api/mutable_resource.go
kl52752: I delete the table. It is incomplete because there are no information about PlaceholderType. Let me know what do you think.

kl52752:
> See comment -- you need to fix the commit message and the PR title.

done

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,24 @@ func (e *ConversionError) Error() string {
...
+//	if the code tries to convert the Resource to the Alpha type,
+//	the operation will fail with this error.
+type UseOfPlaceholderTypeError struct {
+	msg string
+}
+
+func newUseOfPlaceholderTypeError(m string) *UseOfPlaceholderTypeError {
+	return &UseOfPlaceholderTypeError{msg: m}
+}
+func (m *UseOfPlaceholderTypeError) Error() string {
```
bowei, line 71: don't use pointer here...

pkg/cloud/api/mutable_resource.go
```diff
@@ -180,6 +202,7 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
 	if err != nil {
 		return err
 	}
+	// TODOD(kl52752) Add validation that GA is a subset of Beta and Alpha.
```
bowei, line 205: I'm assuming the PR for this is coming very soon?

pkg/cloud/api/mutable_resource.go
```diff
@@ -168,6 +186,10 @@ type mutableResource[GA any, Alpha any, Beta any] struct {
 }
 
 func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
+	if isPlaceholderType(u.ga) {
+		return fmt.Errorf("GA cannot be PlaceholderType")
```
bowei, line 190: Type errors should always print out type and any other information that will be useful for debugging.

fmt.Errorf("GA cannot be PlaceholderType (type is %T)", u)

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,24 @@ func (e *ConversionError) Error() string {
 	return fmt.Sprintf("ConversionError: missing fields %v", e.MissingFields)
 }
 
+// UseOfPlaceholderTypeError is raised when code attempts to convert or operate
```
bowei, line 57: do we need to export this?

bowei: if we do not, please make not public

pkg/cloud/api/mutable_resource.go
```diff
@@ -54,6 +54,24 @@ func (e *ConversionError) Error() string {
...
+//
+//	 Resource[ga.Res, /*alpha*/ PlaceholderType, beta.Res]
+//
+//	if the code tries to convert the Resource to the Alpha type,
+//	the operation will fail with this error.
+type UseOfPlaceholderTypeError struct {
+	msg string
+}
+
+func newUseOfPlaceholderTypeError(m string) *UseOfPlaceholderTypeError {
```
bowei, line 68: probably overkill to have this...

pkg/cloud/api/resource_test.go
```diff
@@ -862,6 +879,73 @@ func TestResourceImpliedVersion(t *testing.T) {
 	}
 }
 
+func TestImpliedVersionForPlaceHolderType(t *testing.T) {
```
bowei, line 882: Inline the struct typedef (copy them) instead of creating a dependency between the two tests (TestImpliedVersion and TestImpliedVersionForPlaceholderType)

e.g. add 

```
type st struct {
		I               int
		NullFields      []string
		ForceSendFields []string
	}
	type stA struct {
		I               int
		A               int
		NullFields      []string
		ForceSendFields []string
	}
	type stB struct {
		I               int
		B               int
		NullFields      []string
		ForceSendFields []string
	}
```

to here and revert the delete.

pkg/cloud/api/resource_test.go
```diff
@@ -862,6 +879,73 @@ func TestResourceImpliedVersion(t *testing.T) {
...
+	PlaceholderTypeBeta := func() (meta.Version, error) {
```
bowei, line 883: I find this func definition indirection and naming to be really hard to understand. because the names don't mean anything. 

It also establishes a pattern that will be hard to maintain, which is that each new test case will need to have a func defined here and then the func referenced below in the test case itself. This makes it so the person writing the test will need to think of a name for the func and then also makes sure the test case matches the set of funcs at the top.

It's better to inline the function -- the name isn't very meaningful and I am also looking around to see if it is used in other places.

Specifically:

```
type iv interface { ImpliedVersion() (meta.Version, error) }

tcs := []struct {
		desc    string
		makeResource func() iv
		wantErr bool
		wantVer meta.Version
	}{
		{
			desc:    "imply ga, PlaceholderType alpha",
			test:    func() iv { return newTestResource[..., ..., ...](nil) }
			wantVer: meta.VersionGA,
		},
```

bowei:
This is close, please address the comments.

pkg/cloud/api/mutable_resource.go
kl52752: doone

pkg/cloud/api/mutable_resource.go
kl52752: removed

pkg/cloud/api/mutable_resource.go
kl52752: done

pkg/cloud/api/mutable_resource.go
kl52752: but time is included in error :) 
maybe sth like fmt.Errorf("GA has unsupported type (type is %T)", u)

pkg/cloud/api/mutable_resource.go
kl52752: yes it's next in line

pkg/cloud/api/resource_test.go
kl52752: done

pkg/cloud/api/resource_test.go
kl52752: done

pkg/cloud/api/mutable_resource.go
kl52752: I added the table in comment

pkg/cloud/api/mutable_resource.go
bowei: Ok -- can you make the change in a follow up PR? I am going to merge this.

PR #166, opened 2023-12-27
Add check that Alpha and Beta are subsets of GA fields

kl52752:
Extend CheckSchema function with Assertion that Alpha and Beta types are subsets of GA.

### DIFF
--- a/pkg/cloud/api/check.go
+++ b/pkg/cloud/api/check.go
@@ -171,3 +171,44 @@ func checkSchema(t reflect.Type) error {
 
 	return nil
 }
+
+// CheckStructuralSubset checks if type From is the subset of To type. Type is a
+// subset of another if it contains fields with the same type and name. This
+// function is not symmetric. Path parameter is used for better error reporting.
+func CheckStructuralSubset(p Path, from, to reflect.Type) error {
+	if from.Kind() != to.Kind() {
+		return fmt.Errorf("%s has different type: %v != %v", p, from.Kind(), to.Kind())
+	}
+	if isBasicT(from) {
+		return nil
+	}
+	switch from.Kind() {
+	case reflect.Pointer:
+		return CheckStructuralSubset(p.Pointer(), from.Elem(), to.Elem())
+
+	case reflect.Struct:
+		for i := 0; i < from.NumField(); i++ {
+			af := from.Field(i)
+			bf, exist := to.FieldByName(af.Name)
+			if !exist {
+				return fmt.Errorf("%s: type %T does not have field %v", p.String(), to, af.Name)
+			}
+			if err := CheckStructuralSubset(p.Field(af.Name), af.Type, bf.Type); err != nil {
+				return err
+			}
+		}
+		return nil
+
+	case reflect.Slice, reflect.Array:
+		return CheckStructuralSubset(p.AnySliceIndex(), from.Elem(), to.Elem())
+
+	case reflect.Map:
+		path := p.AnyMapIndex()
+		err := CheckStructuralSubset(path, from.Key(), to.Key())
+		if err != nil {
+			return err
+		}
+		return CheckStructuralSubset(path, from.Elem(), to.Elem())
+	}
+	return fmt.Errorf("%s Unsupported type %v", p.String(), from.Kind())
+}
--- a/pkg/cloud/api/check_test.go
+++ b/pkg/cloud/api/check_test.go
@@ -19,6 +19,8 @@ package api
 import (
 	"reflect"
 	"testing"
+
+	teststruct "github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/api/converter_test_types"
 )
 
 func TestCheckFieldsAreSet(t *testing.T) {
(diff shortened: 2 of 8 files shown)


pkg/cloud/api/diff.go
```diff
@@ -54,10 +66,10 @@ func (r *DiffResult) add(state DiffItemState, p Path, a, b reflect.Value) {
 		State: state,
 		Path:  p,
 	}
-	if a.IsValid() {
+	if a.IsValid() && a.CanInterface() {
```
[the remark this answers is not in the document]
kl52752: Interface() will panic if it is called on unexported struct's fields https://pkg.go.dev/reflect#Value.Interface.
Our struct has all fields exported but I thought that still we should check that.
I revisited this function and this code is only needed for printing the diff so I changed Interface() to String(). This way we will have the same result (printing which value diff) and it will be safe for both exported and unexported fields.
I added new test for struct with unexported fields.

pkg/cloud/api/diff.go
```diff
@@ -41,6 +41,18 @@ func diff[T any](a, b *T, trait *FieldTraits) (*DiffResult, error) {
 	return d.result, nil
 }
 
+func diffBetweenTypes[T any, V any](a *T, b *V) (*DiffResult, error) {
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/api/diff.go
```diff
@@ -124,18 +136,22 @@ func (d *differ[T]) do(p Path, av, bv reflect.Value) error {
...
 			}
 
+			fp := p.Field(aft.Name)
 			switch d.traits.fieldType(fp) {
 			case FieldTypeOutputOnly, FieldTypeSystem:
 				continue
 			}
 
+			bfv := bv.FieldByName(aft.Name)
+			if !bfv.IsValid() {
```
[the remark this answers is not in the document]
kl52752: we are iterating here only for A struct `for i := 0; i < av.NumField(); i++ {` and we check if fields in A are present in B (it might be in different order but we need the same field name and type). We don't do the same thing the other way. There is a comment for diff function `the behavior of this is not symmetric -- diff(A,B) != diff(B,A).` 
I thought that this is the behavior we want.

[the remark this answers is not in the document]
kl52752: This check is used to detect the `subset of` and I added test cases for schema validation to check if structs are subset of GA in file pkg/cloud/api/resource_test.go.
Do you want me to add specific test for `diffBetweenTypes`?

pkg/cloud/api/diff.go
```diff
@@ -41,6 +41,18 @@ func diff[T any](a, b *T, trait *FieldTraits) (*DiffResult, error) {
 	return d.result, nil
 }
 
+func diffBetweenTypes[A any, B any](a *A, b *B) (*DiffResult, error) {
```
bowei, line 44: This isn't a diff between types, it's a diff between values.

let's call this `diffStructs`

pkg/cloud/api/diff.go
```diff
@@ -55,10 +67,11 @@ func (r *DiffResult) add(state DiffItemState, p Path, a, b reflect.Value) {
 		Path:  p,
 	}
 	if a.IsValid() {
-		di.A = a.Interface()
+		di.A = a.String()
```
bowei, line 70: Why was the previous code incorrect? Are we seeing panics here?

From https://pkg.go.dev/reflect#Value.Interface

```
func (v [Value](https://pkg.go.dev/reflect#Value)) Interface() (i [any](https://pkg.go.dev/builtin#any))
Interface returns v's current value as an interface{}. It is equivalent to:

var i interface{} = (v's underlying value)
It panics if the Value was obtained by accessing unexported struct fields.
```

pkg/cloud/api/diff.go
```diff
@@ -55,10 +67,11 @@ func (r *DiffResult) add(state DiffItemState, p Path, a, b reflect.Value) {
...
+
 	}
 	if b.IsValid() {
-		di.B = b.Interface()
+		di.B = b.String()
```
bowei, line 74: See comment above, I'm not sure this should be changed.

pkg/cloud/api/mutable_resource.go
```diff
@@ -195,11 +195,41 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
...
+	if isPlaceholderType(t2) {
+		return nil
+	}
+	diff, err := diffBetweenTypes(t1, t2)
+	if err != nil {
+		return err
+	}
+	for _, d := range diff.Items {
+		if d.State == DiffItemOnlyInA {
+			return fmt.Errorf("GA fields not present in Alpha type. Paths: %s", d.Path)
```
bowei, line 230: error message doesn't match func 

fmt.Errorf("checkSubsetOf(): field %s is in %T but not in %T",...)

pkg/cloud/api/mutable_resource.go
```diff
@@ -195,11 +195,41 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
 	if err != nil {
 		return err
 	}
+	ga, _ := u.ToGA()
+	alpha, _ := u.ToAlpha()
+	err = checkSubsetOf(ga, alpha)
+
+	if err != nil {
+		return fmt.Errorf("checkSubsetOf(ga, alpha) = %v, want nil", err)
```
bowei, line 203: You should always print out the types %T so people have a straightforward time debugging this when it doesn't pass instead of just an error message with no context.

pkg/cloud/api/mutable_resource.go
```diff
@@ -195,11 +195,41 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
 	if err != nil {
 		return err
 	}
+	ga, _ := u.ToGA()
+	alpha, _ := u.ToAlpha()
+	err = checkSubsetOf(ga, alpha)
```
bowei, line 200: Don't you need to `fill` the struct first? otherwise this will ignore nil sub structs? Add test cases for that if they don't exist.

pkg/cloud/api/mutable_resource.go
```diff
@@ -195,11 +195,41 @@ func (u *mutableResource[GA, Alpha, Beta]) CheckSchema() error {
...
+
+	if err != nil {
+		return fmt.Errorf("checkSubsetOf(ga, beta) = %v, want nil", err)
+	}
+
+	return nil
+}
+
+func checkSubsetOf[T1 any, T2 any](t1 *T1, t2 *T2) error {
+	if isPlaceholderType(t2) {
```
bowei, line 221: Don't put this logic here. It's confusing and technically placeholder type is not a superset of any struct except the empty struct. This will lead to bugs later because the special case behavior is not consistent.

pkg/cloud/gen.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 Google LLC
+Copyright 2024 Google LLC
```
bowei, line 2: This should go in a different commit.

pkg/cloud/api/diff.go
kl52752: We don't see panic here because type we use for rGraph does not have unexported fields. But don't we want to support that case?

pkg/cloud/api/mutable_resource.go
kl52752: in checkSubsetOf I leverage diff function to reduce code duplication but in fact what diff does is checking fields type and if it match then it comparing values. If I have unset field and they are not present in type B it will be cache in https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/166/files/271e930ec4be6ca5aea04bf26a9a2077176d4195. Structs are treated as any other fields here. 
I added test case for struct.

pkg/cloud/api/mutable_resource.go
kl52752: done

pkg/cloud/api/diff.go
kl52752: done

pkg/cloud/api/mutable_resource.go
kl52752: done

pkg/cloud/gen.go
kl52752: it is auto generated in `make` - I removed this form commit

pkg/cloud/api/mutable_resource.go
kl52752: done

pkg/cloud/api/diff.go
bowei: It's not very obvious this was your intent for the change.

1. You should not convert to String(). This violates the contract -- the type in the `di` struct is a `interface{}` and you are silently stringifying it in this layer.
2. If the check is to skip unexported fields, you need to put a comment here as it is not obvious that was the intent.
3. You need to add a test case that exercises the path.

pkg/cloud/api/mutable_resource.go
bowei: The link you gave doesn't work. What I'm asking about is this:

```
type A struct {
  B *B
}

type B struct {
  ...
  C *C
}

a := &A{}
```

if `a.B == nil`, does this code go and check the schema of the fields below, e.g. `a.B.C`

kl52752:
cc/ @tangenti

pkg/cloud/api/diff.go
kl52752: I added comment and if for unexported type to convert to string so in this case we will have some error reporting.

pkg/cloud/api/diff.go
kl52752: as above

pkg/cloud/api/mutable_resource.go
kl52752: you are right, this scenario was not working. I decided to create a new function based on reflect.Type there is no benefit from sharing the diff function in this case.

kl52752:
cc/ @mag-kol

pkg/cloud/api/converter_test_types/converter-structs.go
```diff
@@ -0,0 +1,52 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+// This file contains structs for testing conversion. Type to test conversion
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,41 @@ func checkSchema(t reflect.Type) error {
 
 	return nil
 }
+
+// canConvertToT checks if we can convert type From to type To. Conversion is
```
bowei, line 175: I find this confusing -- what is the purpose of the path `p` here? It is not mentioned at all in the docstring.

kl52752: This is only for better error reporting. This way we have nested fields info and without the path we can only log field name and type but if this is in nested structs we won't have this information

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,41 @@ func checkSchema(t reflect.Type) error {
...
+// possible if type From has the subset of fields with the same type and name as
+// in To type. This function is not symmetric.
+func canConvertToT(p Path, from, to reflect.Type) error {
```
tangenti, line 178: Should this be rename to something similar to `CheckConversion`? 

So the functions of `CheckXXX` would return an error, and `CanXXX` would return a boolean (e.g. `CanInterface`)

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,41 @@ func checkSchema(t reflect.Type) error {
...
+		return canConvertToT(p.Pointer(), from.Elem(), to.Elem())
+
+	case reflect.Struct:
+		for i := 0; i < from.NumField(); i++ {
+			af := from.Field(i)
+			bf, exist := to.FieldByName(af.Name)
+			if !exist {
+				return fmt.Errorf("%s: type %T does not have field %v", p.String(), to, af.Name)
+			}
+			if err := canConvertToT(p.Field(af.Name), af.Type, bf.Type); err != nil {
```
tangenti, line 193: Is there any GCP API struct referring to the pointer of itself?

pkg/cloud/api/diff.go
```diff
@@ -55,10 +67,22 @@ func (r *DiffResult) add(state DiffItemState, p Path, a, b reflect.Value) {
...
+		} else {
+			di.A = a.String()
+		}
 	}
 	if b.IsValid() {
-		di.B = b.Interface()
+		// Interface() will panic if is called on unexported types in this case
+		// the best we can do is to pass its name to the result.
+		if a.CanInterface() {
+			di.A = a.Interface()
```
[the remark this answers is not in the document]
kl52752: yes, fixed

pkg/cloud/api/check.go
kl52752: done

pkg/cloud/api/check.go
kl52752: good question :) not the ones we use in rGraph right now

pkg/cloud/api/check.go
kl52752: Added comment about path parameter. Does it make sense?

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,42 @@ func checkSchema(t reflect.Type) error {
...
+	case reflect.Slice:
+		return CheckConversion(p.AnySliceIndex(), from.Elem(), to.Elem())
+
+	case reflect.Map:
+		path := p.AnyMapIndex()
+		err := CheckConversion(path, from.Key(), to.Key())
+		if err != nil {
+			return err
+		}
+		return CheckConversion(path, from.Elem(), to.Elem())
```
bowei, line 209: we should document

We should add all of the types:

```
case reflect.Channel:
  // this is an error...
default:
  // Assume 
```

pkg/cloud/api/check_test.go
```diff
@@ -272,3 +274,155 @@ func TestCheckSchema(t *testing.T) {
 		})
 	}
 }
+
+func TestConvertToT(t *testing.T) {
```
bowei, line 278: we need 

* slice -> pointer
* slice -> slice 
* map[] -> pointer

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,42 @@ func checkSchema(t reflect.Type) error {
 
 	return nil
 }
+
+// CheckConversion checks if we can convert type From to type To. Conversion is
+// possible if type From has the subset of fields with the same type and name as
+// in To type. This function is not symmetric. Path parameter is used for better
+// error reporting.
+func CheckConversion(p Path, from, to reflect.Type) error {
```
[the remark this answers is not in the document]
kl52752: renamed to `CheckStructuralSubset` because as Hugo noticed isXX should return bool and checkXX error. I want to return error here so we have the information which field is missing in subset.

pkg/cloud/api/check_test.go
kl52752: done

pkg/cloud/api/check.go
kl52752: added check for Basic type and error on not supported (others)

bowei (commented):
minor things, otherwise lgtm

bowei:
ping, this PR is close.

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,46 @@ func checkSchema(t reflect.Type) error {
 
 	return nil
 }
+
+// CheckStructuralSubset checks if type From is the subset of To type. Type is a
+// subset of another if it contains fields with the same type and name. This
+// function is not symmetric. Path parameter is used for better error reporting.
+func CheckStructuralSubset(p Path, from, to reflect.Type) error {
```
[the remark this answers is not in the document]
kl52752: I will stick with this name. As Hugo mentioned function starting with `Is` should return bool and I want to return error here and pass the path which is missing

kl52752:
> ping, this PR is close.

Do you want me to rebase this PR to master branch?

bowei:
Yes, rebase to master.

kl52752:
> Yes, rebase to master.

master branch is missing few commits from rgraph-unstable,. Can we merge this PR to rgraph and then I will prepare merge rgraph to master.

pkg/cloud/api/check.go
```diff
@@ -171,3 +171,44 @@ func checkSchema(t reflect.Type) error {
...
```
bowei, line 178: CheckStructuralSubset(from, to reflect.Type) calls

  checkStructuralSubsetImpl(Path{})

checkStructuralSubsetImpl(p Path, from, to)

PR #168, opened 2024-01-19
Wrap coverage run in shell, fix inconsistency in results

bowei:
- Wrap the coverage runner in a script for easier maintenance
- Running coverage over all of the packages in a single invocation seems to not give the same coverage numbers as running it on a single package basis. Run the coverage tool one package at a time for now.
- Add -packagePrefix to make the coverage file less verbose

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,6 @@ build: gen
 	go build ./...
 	mkdir -p bin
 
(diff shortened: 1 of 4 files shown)


swetharepakula (commented):
/lgtm

PR #170, opened 2024-01-25
Add bosko project handler for e2e tests.

kl52752:


aojea:
you should use the trick of of a tools.go file to not bring the dependencies to the project and only use them for testing

https://github.com/kubernetes/kubernetes/commit/f981628ce897a96a8a08a2a21c4d6bdb41601fbd
https://github.com/kubernetes/ingress-gce/tree/master/hack/tools

e2e/project_holder.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+	boskosclient "sigs.k8s.io/boskos/client"
+	"sigs.k8s.io/boskos/common"
```
aojea, line 26: I see, so this is the file that is pulling all the dependencies, do you have to embed boskos client here

Are there any other options?

@justinsb @bentheelder do you know if we can avoid pulling all these dependencies?

bowei:
Can we fix the dependencies issue? This is pulling in a huge number of files.

e2e/project_holder.go
aojea: >  Can we fix the dependencies issue? This is pulling in a huge number of files.

@bowei this is the thread about it :)

bowei:
Ping on this change...

bowei:
Ping.

kl52752:
Bosko project is not needed anymore for setting up the e2e test grid. Closing this PR

PR #173, opened 2024-02-29
Deep copy path in Differ Results

kl52752:
Differ returns a Results list of all paths that mismatch. The list is a pair of paths and status.
The path is of type slice of strings.
We need to deep copy the path otherwise it will be overridden in the next iteration and as a result we will get all paths pointing to the same value.


kl52752:
cc/ @AwesomePatrol

kl52752:
assign/ @bowei

bowei:
Commit comments need to be more descriptive

One line summary

Longer description.

Please follow format from https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

kl52752:
Updated @bowei PTAL

bowei:
You added some extra whitespace -- please remove next time.

/lgtm
/approve

PR #176, opened 2024-03-19
Plan BackendService recreate on Network change

kl52752:
When Network property in compute.BackendService change planing should return Recreate action.

### DIFF
--- a/pkg/cloud/rgraph/rnode/backendservice/backendservice_test.go
+++ b/pkg/cloud/rgraph/rnode/backendservice/backendservice_test.go
@@ -18,6 +18,7 @@ package backendservice
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
@@ -28,7 +29,10 @@ import (
 	"google.golang.org/api/compute/v1"
 )
 
-const proj = "proj-1"
+const (
+	proj       = "proj-1"
+	hcSelfLink = "https://www.googleapis.com/compute/v1/projects/proj-1/global/healthChecks/hcName"
+)
 
 func TestBackendServiceSchema(t *testing.T) {
 	key := meta.GlobalKey("key-1")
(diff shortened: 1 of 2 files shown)


pkg/cloud/rgraph/rnode/backendservice/node.go
```diff
@@ -38,6 +38,8 @@ var _ rnode.Node = (*backendServiceNode)(nil)
 
 func (n *backendServiceNode) Resource() rnode.UntypedResource { return n.resource }
 
+func (n *backendServiceNode) BackendResource() BackendService { return n.resource }
```
bowei, line 41: Two things:

1. Why not cast Resource() to the concrete type in the test? Why do we have to wrap up the cast with a helper method? If this is just for testing, I think it doesn't make sense to do this...

2. If this is a pattern we want to follow, I recommend naming this `TypedResource()` and implementing it on every resource, not just the BackendService.

We should always make the code as uniform as possible -- unfortunately Golang does not allow us to implement these kinds of uniformity constraints in the type system itself, but we should always be looking for ways to make things the same across _all_ of the types to make it easier for the user of the library to work with. 

This includes patterns such as this.

kl52752: sorry this is a leftover form other tests it should not be in this review

pkg/cloud/rgraph/rnode/backendservice/backendservice_test.go
```diff
@@ -59,11 +57,32 @@ func TestActionUpdate(t *testing.T) {
...
+
+	gotNode, err := createBackendServiceNode("bs-name", modify)
+	if err != nil {
+		t.Fatalf("createBackendServiceNode(bs-name, _) = %v, want nil", err)
+	}
+	gotBs := gotNode.resource
+
+	_, err = gotBs.ToGA()
+	if err != nil {
+		t.Errorf("error %v", err)
```
bowei, line 81: t.Errorf("gotBs.ToGA() = %v, want nil", err)

pkg/cloud/rgraph/rnode/backendservice/node.go
```diff
@@ -69,9 +69,9 @@ func (n *backendServiceNode) Diff(gotNode rnode.Node) (*rnode.PlanDetails, error
 	}
 
 	for _, delta := range diff.Items {
-		//TODO(kl52752) explore which fields can't be updated
 		switch {
-		case delta.Path.Equal(api.Path{}.Pointer().Field("LoadBalancingScheme")):
+		case delta.Path.Equal(api.Path{}.Pointer().Field("LoadBalancingScheme")),
```
bowei, line 73: Let's add a brief comment here to help future maintainers reason about the behavior:

```
// These fields cannot be changed in place and require the
// resource to be recreated.
```

bowei (commented):
Some minor comments.

pkg/cloud/rgraph/rnode/backendservice/node.go
kl52752: done

pkg/cloud/rgraph/rnode/backendservice/backendservice_test.go
kl52752: done

PR #177, opened 2024-03-20
Fix code gen for networkservices Meshes API

kl52752:


pkg/cloud/gen/main.go
```diff
@@ -309,8 +309,13 @@ func (m *Mock{{.Service}}Obj) ToGA() *{{.GA.FQObjectType}} {
 	}
 }
 
+func shouldAddIdToCreateCall(obj string) bool {
```
bowei, line 312: Can we document this to make it clear to future maintainers?

I also suggest changing the name to make it easier to understand.

```
// callOperationRequiresID returns true if the ServiceInfo.Object is
// of a particular type.
func callOperationRequiresID(objectName string) bool 
```

kl52752: done, I added couple more types I found

pkg/cloud/gen/main.go
```diff
@@ -309,8 +309,19 @@ func (m *Mock{{.Service}}Obj) ToGA() *{{.GA.FQObjectType}} {
 	}
 }
 
+// callOperationRequiresID returns true if the ServiceInfo.Object is
+// of a particular type.
+func callOperationRequiresID(obj string) bool {
+	return strings.HasSuffix(obj, "Route") ||
+		strings.HasSuffix(obj, "Mesh") ||
+		strings.HasSuffix(obj, "tlsRoute") ||
```
bowei, line 317: Do we need HasSuffix or can do a full string match?

My concern is that this may inadvertently cause types to be included that we weren't aware of.

kl52752: in this case this is only for Routes, we have different type (tcp, tlc etc) but we can list them all

kl52752: changed

PR #179, opened 2024-03-22
implement Action Update for AppNet TCPRoute

kl52752:
* Implement action update using Patch method for AppNet TCPRoute
* Extract NotFoundError to new cerrors module 
* Add e2e tests

### DIFF
--- a/e2e/rgraph_update_action_test.go
+++ b/e2e/rgraph_update_action_test.go
@@ -17,7 +17,6 @@ package e2e
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"testing"
 
@@ -39,8 +38,12 @@ func buildBackendService(graphBuilder *rgraph.Builder, name string, hcID *cloud.
 	bsMutResource.Access(func(x *compute.BackendService) {
 		x.LoadBalancingScheme = "INTERNAL_SELF_MANAGED"
 		x.Protocol = "TCP"
+		x.PortName = "http"
+		x.SessionAffinity = "NONE"
 		x.Port = port
+		x.TimeoutSec = 30
 		x.HealthChecks = []string{hcID.SelfLink(meta.VersionGA)}
+		x.ConnectionDraining = &compute.ConnectionDraining{}
 	})
 	bsResource, err := bsMutResource.Freeze()
 	if err != nil {
@@ -64,9 +67,11 @@ func buildHealthCheck(graphBuilder *rgraph.Builder, name string, checkIntervalSe
 		x.HealthyThreshold = 5
 		x.TimeoutSec = 6
 		x.Type = "HTTP"
+		x.UnhealthyThreshold = 2
 		x.HttpHealthCheck = &compute.HTTPHealthCheck{
 			RequestPath: "/",
 			Port:        int64(9376),
+			ProxyHeader: "NONE",
 		}
 	})
 	hcRes, err := hcMutRes.Freeze()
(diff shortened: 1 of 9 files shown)


pkg/cloud/utils.go
```diff
@@ -18,11 +18,13 @@ package cloud
 
 import (
 	"encoding/json"
+	"errors"
```
bowei, line 21: suggest putting this in its own package.

`pkg/cloud/cerrors/errors.go`

```
package cerrors

func IsGoogleAPINotFound(err error) bool {...}
```

Then the usage will be:

```
cerrors.IsGoogleAPINotFound(err)
```

which should be self documenting.

e2e/test-helpers.go
```diff
@@ -0,0 +1,127 @@
...
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/exec"
+)
+
+func expectActions(got []exec.Action, want []exec.ActionType) error {
```
bowei, line 30: // expectActions compares got with want, checking that the list of actions is <???>

What is the predicate that is being checked here?

You need to document all functions that are utility and intended to be used in a common way, even if they are not exported.

This is especially important as the docstring should define a loose sense of what you want the contract of the function to be. Otherwise, people maintaining the code will not understand what you intended and may not make the right choices.

* Does order matter?
* Are there fields that are being ignored? If this is temporary situation, then document a TODO.

e2e/test-helpers.go
```diff
@@ -0,0 +1,127 @@
...
+	var errs []error
+WantLoop:
+	for _, wantT := range want {
```
bowei, line 33: If you do this, if there are extra in `got`, you wouldn't know. This is a bug.

If order does not matter, I suggest putting into a set or map and then calling `cmp.Diff` instead of trying to write custom loops like this.

pkg/cloud/utils.go
kl52752: done

e2e/rgraph_update_action_test.go
```diff
@@ -147,8 +151,8 @@ func TestHcUpdateWithBackendService(t *testing.T) {
 			t.Logf("delete health check: %v", err)
 		}
 	})
-	checkGCEHealthCheck(t, ctx, hcID, 15)
-	checkGCEBackendService(t, ctx, hcID, bsID, 80)
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/utils.go
```diff
@@ -337,3 +339,13 @@ func aggregatedListKey(k *meta.Key) string {
 		return "unknownScope"
 	}
 }
+
+func isErrorCode(err error, code int) bool {
```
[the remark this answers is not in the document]
kl52752: done

e2e/test-helpers.go
kl52752: the idea was to check if got contains want not if they are equal. 
But actually we can check all actions

kl52752: changed, I needed to update existing test and add all actions including events

e2e/test-helpers.go
kl52752: done

bowei:
Please clean up the commits:

You have changes to the main logic, add errors.go etc but the commit description is "Create e2e tests for TCPRoute Update Action" with no description.

Split the commit into pieces that are more self-contained and update the one line message and the description appropriately.

It is important that we keep the commits in the codebase meaningful as it is a form of documentation as well for the history of the changes.

bowei:
Thanks for the revisions.

/lgtm
/approve

PR #180, opened 2024-03-26
Add health check traits

kl52752:


pkg/cloud/rgraph/rnode/healthcheck/type_trait.go
```diff
@@ -41,7 +41,48 @@ func (*typeTrait) FieldTraits(meta.Version) *api.FieldTraits {
...
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("HttpHealthCheck").Pointer().Field("Response"))
+
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("HttpsHealthCheck"))
+	// This field is not supported
+	dt.OutputOnly(api.Path{}.Pointer().Field("HttpsHealthCheck").Pointer().Field("PortName"))
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("HttpsHealthCheck").Pointer().Field("Response"))
+
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("SslHealthCheck"))
+	// This field is not supported
+	dt.OutputOnly(api.Path{}.Pointer().Field("SslHealthCheck").Pointer().Field("PortName"))
```
AwesomePatrol, line 70: (in separate patch) should we consider a new type - `Unsupported`? Not sure if `Output` is the best choice here, even `System` looks better. API description also doesn't help - is it deprecated or a planned feature?

pkg/cloud/rgraph/rnode/healthcheck/type_trait.go
```diff
@@ -29,7 +29,7 @@ type typeTrait struct {
 	api.BaseTypeTrait[compute.HealthCheck, alpha.HealthCheck, beta.HealthCheck]
 }
 
-func (*typeTrait) FieldTraits(meta.Version) *api.FieldTraits {
+func (*typeTrait) FieldTraits(v meta.Version) *api.FieldTraits {
 	dt := api.NewFieldTraits()
 	// Built-ins
 	dt.OutputOnly(api.Path{}.Pointer().Field("Fingerprint"))
```
AwesomePatrol, line 35: In `urlmap` it is set to `System`. Which way is correct?

pkg/cloud/rgraph/rnode/healthcheck/healthcheck_test.go
```diff
@@ -20,13 +20,319 @@ import (
...
+		t.Fatalf("hcMutRes.Freeze(_) = %v, want nil", err)
+	}
+	b := NewBuilderWithResource(hcRes)
+	alphaNode, err := b.Build()
+	if err != nil {
+		t.Fatalf("b.Build() = %v, want nil", err)
+	}
+	plan, err = wantNode.Diff(alphaNode)
+	if err != nil || plan.Diff == nil {
+		t.Fatalf("wantNode.Diff(wantNode) = (%v, %v), want (diff, nil)", plan.Diff, err)
```
AwesomePatrol, line 267: nit: plan can be nil if an error occurred (will get panic instead of nice error)

AwesomePatrol (approved):
lgtm. I added a few questions in the comments.

Do you plan to add e2e tests calling real GCP API to verify that there are no other fields defaulting? It would be great to avoid/verify PRs like #181

pkg/cloud/rgraph/rnode/healthcheck/type_trait.go
kl52752: This field is Deprecated but still this value might be already set and in this case it will be returned. OutputOnly indicates that this field might be returned by the cloud but should not be set by clients

pkg/cloud/rgraph/rnode/healthcheck/healthcheck_test.go
kl52752: if error occurs then `err != nil` will return true and `plan.Diff == nil` won't be executed, right?

pkg/cloud/rgraph/rnode/healthcheck/type_trait.go
kl52752: I just noticed that this is an error because HC does not support fingerprint

pkg/cloud/rgraph/rnode/healthcheck/healthcheck_test.go
AwesomePatrol: yes, I meant L267

kl52752: oh yes, god catch!

PR #182, opened 2024-03-28
Implement BackendService traits

kl52752:
* Add traits to BackendService for all versions (ga, alpha, beta)
* Increase test coverage for BackendService rNode

### DIFF
--- a/pkg/cloud/rgraph/rnode/backendservice/type_trait.go
+++ b/pkg/cloud/rgraph/rnode/backendservice/type_trait.go
@@ -29,7 +29,7 @@ type typeTrait struct {
 	api.BaseTypeTrait[compute.BackendService, alpha.BackendService, beta.BackendService]
 }
 
-func (*typeTrait) FieldTraits(meta.Version) *api.FieldTraits {
+func (*typeTrait) FieldTraits(v meta.Version) *api.FieldTraits {
 	dt := api.NewFieldTraits()
 	// Built-ins
 	dt.OutputOnly(api.Path{}.Pointer().Field("Fingerprint"))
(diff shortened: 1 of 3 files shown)


pkg/cloud/rgraph/rnode/backendservice/type_trait.go
```diff
@@ -43,11 +43,136 @@ func (*typeTrait) FieldTraits(meta.Version) *api.FieldTraits {
...
 	dt.OutputOnly(api.Path{}.Pointer().Field("SelfLink"))
 
-	dt.OutputOnly(api.Path{}.Pointer().Field("Iap").Field("Oauth2ClientSecretSha256"))
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("Iap"))
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("Iap").Pointer().Field("Enabled"))
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("Iap").Pointer().Field("Oauth2ClientId"))
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("Iap").Pointer().Field("Oauth2ClientSecret"))
+	dt.OutputOnly(api.Path{}.Pointer().Field("Iap").Pointer().Field("Oauth2ClientSecretSha256"))
+
+	dt.AllowZeroValue(api.Path{}.Pointer().Field("Port"))
```
AwesomePatrol, line 52: Shouldn't it be OutputOnly if it is deprecated?

kl52752: This is a good question :) For me deprecated doesn't mean that you cannot use this but just shouldn't.

bowei: I think deprecated or not should not affect the type trait -- deprecation is says "it still works, but please don't use it" -- but if we need to continue to use it, then that's what it is...

bowei:
This seems to be a huge number of optional values. Are we sure we can't just set the outer struct as optional, but leave the rest of the values required?

kl52752:
> can't just set the outer struct as optional, but leave the rest of the values required?

BackendService is very complex structure. Looking at the `SecuritySettings` field there are 3 fields but each of them has restrictions when can be set and this restriction depends on LoadBalancingScheme.

```
type SecuritySettings struct {
	// AwsV4Authentication: The configuration needed to generate a signature
	// for access to private storage buckets that support AWS's Signature
	// Version 4 for authentication. Allowed only for INTERNET_IP_PORT and
	// INTERNET_FQDN_PORT NEG backends.
	AwsV4Authentication *AWSV4Signature `json:"awsV4Authentication,omitempty"`

	// ClientTlsPolicy: Optional. A URL referring to a
	// networksecurity.ClientTlsPolicy resource that describes how clients
	// should authenticate with this service's backends. clientTlsPolicy
	// only applies to a global BackendService with the loadBalancingScheme
	// set to INTERNAL_SELF_MANAGED. If left blank, communications are not
	// encrypted.
	ClientTlsPolicy string `json:"clientTlsPolicy,omitempty"`

	// SubjectAltNames: Optional. A list of Subject Alternative Names (SANs)
	// that the client verifies during a mutual TLS handshake with an
	// server/endpoint for this BackendService. When the server presents its
	// X.509 certificate to the client, the client inspects the
	// certificate's subjectAltName field. If the field contains one of the
	// specified values, the communication continues. Otherwise, it fails.
	// This additional check enables the client to verify that the server is
	// authorized to run the requested service. Note that the contents of
	// the server certificate's subjectAltName field are configured by the
	// Public Key Infrastructure which provisions server identities. Only
	// applies to a global BackendService with loadBalancingScheme set to
	// INTERNAL_SELF_MANAGED. Only applies when BackendService has an
	// attached clientTlsPolicy with clientCertificate (mTLS mode).
	SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```

And this model when fields value depends on other setting are more.

bowei:
We should file a TODO to track making this less verbose/onerous to deal with.

/lgtm
/approve

PR #183, opened 2024-04-03
Fix TCP Route test

AwesomePatrol:
Previously, the tests relied on hardcoded names of NEGs created by
neg-controller in GKE cluster. Now all e2e tests can be run without GKE.

### DIFF
--- a/e2e/tcproute_test.go
+++ b/e2e/tcproute_test.go
@@ -37,9 +37,8 @@ import (
 
 const (
 	meshName   = "test-mesh"
-	negName    = "k8s1-3894226e-default-foo-backend-80-28ec9cf3"
-	negName2   = "k8s1-54982af5-default-bar-svc-8080-5aabd6b1"
-	zone       = "us-central1-c"
+	region     = "us-central1"
+	zone       = region + "-c"
 	routeCIDR  = "10.240.3.83/32"
 	routeCIRD2 = "10.240.4.83/32"
 )
@@ -114,10 +113,15 @@ func TestTcpRoute(t *testing.T) {
 }
 
 func buildNEG(graphBuilder *rgraph.Builder, name, zone string) (*cloud.ResourceID, error) {
-	negID := networkendpointgroup.ID(testFlags.project, meta.ZonalKey(name, zone))
+	negID := networkendpointgroup.ID(testFlags.project, meta.ZonalKey(resourceName(name), zone))
 	negMut := networkendpointgroup.NewMutableNetworkEndpointGroup(testFlags.project, negID.Key)
 	negMut.Access(func(x *compute.NetworkEndpointGroup) {
 		x.Zone = zone
+		x.NetworkEndpointType = "GCE_VM_IP_PORT"
+		x.Name = negID.Key.Name
+		x.Network = defaultNetworkURL()
+		x.Subnetwork = defaultSubnetworkURL()
+		x.Description = "neg for rGraph test"
 	})
 
 	negRes, err := negMut.Freeze()
@@ -125,7 +129,7 @@ func buildNEG(graphBuilder *rgraph.Builder, name, zone string) (*cloud.ResourceI
 		return nil, err
 	}
 	negBuilder := networkendpointgroup.NewBuilder(negID)
-	negBuilder.SetOwnership(rnode.OwnershipExternal)
+	negBuilder.SetOwnership(rnode.OwnershipManaged)
 	negBuilder.SetState(rnode.NodeExists)
 	negBuilder.SetResource(negRes)
 	graphBuilder.Add(negBuilder)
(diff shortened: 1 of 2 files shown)


e2e/tcproute_test.go
```diff
@@ -362,16 +374,12 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
 		{bsID2, routeCIRD2},
 	}
 	tcprID, err = buildTCPRouteWithBackends(graphBuilder, "tcproute-test", meshURL, routes)
-	t.Cleanup(func() {
```
kl52752, line 365: This was needed because TCPRoute is referencing BackendService and you cannot delete BackendService when it is used by other resource. I added it 2 times because if test failed after TCPRoute route was created and before it was updated it might leaked the resource.  In Update operation we are creating a new BS and referencing it to TCPRoute.

e2e/tcproute_test.go
```diff
@@ -114,18 +113,23 @@ func TestTcpRoute(t *testing.T) {
...
 
 func buildNEG(graphBuilder *rgraph.Builder, name, zone string) (*cloud.ResourceID, error) {
-	negID := networkendpointgroup.ID(testFlags.project, meta.ZonalKey(name, zone))
+	negID := networkendpointgroup.ID(testFlags.project, meta.ZonalKey(resourceName(name), zone))
 	negMut := networkendpointgroup.NewMutableNetworkEndpointGroup(testFlags.project, negID.Key)
 	negMut.Access(func(x *compute.NetworkEndpointGroup) {
 		x.Zone = zone
+		x.NetworkEndpointType = "GCE_VM_IP_PORT"
+		x.Name = negID.Key.Name
+		x.Network = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/global/networks/default", testFlags.project)
```
bowei, line 122: suggest putting in e2e/testutils/testutils.go

```
type Res struct {
  Project, Region, Zone, Resource string
}

// NetworkURL returns a URL to be used for testing.
// If Project is empty, defaults to "test-project".
// If Name is empty, defaults to "default".
func NetworkURL(r *R) string {
  return fmt.Sprintf(...)
}
```

then usage:

```
  x.Network = testutils.NetworkURL(testutils.Res{Project: "xxx"}
```

this way we can 1) make it easy to write tests, 2) reuse

AwesomePatrol: I think `cloud.ResourceID` already covers it, so I added `defaultNetwork` and `defaultSubnetwork`.

AwesomePatrol: I just realized that there is `cloud.NewSubnetworkResourceID`, so it is even simpler (now).

e2e/test-helpers.go
```diff
@@ -127,3 +127,11 @@ Outer:
 		t.Fatalf("Rule ServiceName %s, not found in expected: %v", rule.Action.Destinations[0].ServiceName, svcIds)
 	}
 }
+
+func defaultNetwork() string {
```
bowei, line 131: defaultNetworkURL

e2e/test-helpers.go
```diff
@@ -127,3 +127,11 @@ Outer:
...
+	return cloud.NewNetworksResourceID(testFlags.project, "default").SelfLink(meta.VersionGA)
+}
+
+func defaultSubnetwork() string {
```
bowei, line 135: defaultSubnetworkURL

e2e/tcproute_test.go
```diff
@@ -330,16 +338,20 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, tcprID)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, bsID)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, hcID)},
-		{Type: exec.ActionTypeMeta, Name: eventName(negID)},
+		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, negID)},
```
aojea, line 341: thisis changing the behavior of the test, isn't it?

e2e/tcproute_test.go
```diff
@@ -370,8 +382,8 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
 		{Type: exec.ActionTypeUpdate, Name: actionName(exec.ActionTypeUpdate, tcprID)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, bsID2)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, hcID2)},
+		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, negID2)},
 		{Type: exec.ActionTypeMeta, Name: eventName(negID)},
-		{Type: exec.ActionTypeMeta, Name: eventName(negID2)},
```
aojea, line 374: why this order changeS?

AwesomePatrol: order doesn't matter as it will be compared with maps

e2e/tcproute_test.go
AwesomePatrol: yes, previously there were hardcoded references to existing NEGs. Now they are created as needed for the duration of the test

kl52752: that's a good point, you don't have to change the behaviour of the test. Just create the NEG with raw API call and leave the node as it was (External).

PR #184, opened 2024-04-05
refactor executor

kl52752:
Change Executor `Run` function to accept as an input list of pending actions. This way we can implement retires and run Executor multiple times with different set of actions.

Change SerialExecutor constructor to accept cloud client as an input.

### DIFF
--- a/cmd/resgraph/viz/main.go
+++ b/cmd/resgraph/viz/main.go
@@ -154,13 +154,13 @@ func showStep(w http.ResponseWriter, cl cloud.Cloud, idx int, step *testlib.Step
 	outln("")
 
 	var viz exec.GraphvizTracer
-	ex, err := exec.NewSerialExecutor(result.Actions, exec.DryRunOption(false), exec.TracerOption(&viz))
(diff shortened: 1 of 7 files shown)


bowei:
Why not create a new Executor to continue execution? It makes the contract and lifecycle of the object more complex to be able to call into the object multiple times after a "done" state.

kl52752:
> Why not create a new Executor to continue execution? It makes the contract and lifecycle of the object more complex to be able to call into the object multiple times after a "done" state.

Everything depends on the retry strategy. Please take a look at the doc.

If we will implement retries inside the executor using provided by clients function then the approach of creating the executor per run is ok. Error strategy is not enough in this case because we do not retry failed operation, we can ignore errors only.  

If we will not do the retries in the executor then IMO it is better to add ability to the clients to retry the operation on the same executor if needed.

bowei:
I don't quite understand your comment. Seems like to make it possible rerun `Action`, you need to reset the state of the Action, as internally it has a PendingEvents list that will not be in the right state after running the executor...

kl52752:
Discussed offline that this refactor is not needed

PR #186, opened 2024-04-11
implement Parallel executor

kl52752:
Implement parallel executor based on parallel queue

This change is implemented on top of `refactor executor` [PR](https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/184)

In this change other changes were introduced:
* Add Stop on done configuration to parallel queue
* Add Timeout option to executor config


bowei:
Please make sure your commit messages are the right format and are descriptive.

kl52752:
> Please make sure your commit messages are the right format and are descriptive.

This PR is created based on `refactor executor`. I will fix the message once we decide on refactor strategy :)

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+
+}
+
+func (ex *parallelExecutor) executeTasks() {
+	klog.V(4).Infof("executeTasks() pending: %d", len(ex.result.Pending))
+
+	ex.lock.Lock()
+	defer ex.lock.Unlock()
+
+	toProcess := []Action{}
```
bowei, line 129:
```
var (
  taskWasRun bool
  toProcess []Action
)
```

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, a := range ex.result.Pending {
+		if a.CanRun() {
+			klog.V(4).Infof("run task: %s", a)
+			ex.pq.Add(a)
```
bowei, line 133: taskWasRun = true

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+		} else {
+			// store actions that cannot be run in a new array
+			toProcess = append(toProcess, a)
+		}
+	}
+	klog.V(4).Infof("remaining pending actions: %d", len(toProcess))
+	// update Pending array only if actions were run
+	if len(toProcess) < len(ex.result.Pending) {
```
bowei, line 141: if taskWasRun {

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, a := range ex.result.Pending {
+		if a.CanRun() {
+			klog.V(4).Infof("run task: %s", a)
+			ex.pq.Add(a)
+		} else {
+			// store actions that cannot be run in a new array
+			toProcess = append(toProcess, a)
+		}
+	}
+	klog.V(4).Infof("remaining pending actions: %d", len(toProcess))
```
bowei, line 139: Log messages should start with capitalized

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	if len(ex.result.Errors) > 0 || len(ex.result.Pending) != 0 {
+		return ex.result, fmt.Errorf("Executor did not process all actions")
+	}
+	return ex.result, nil
+
+}
+
+func (ex *parallelExecutor) executeTasks() {
```
bowei, line 123: executeTasks => queueRunnableActions

pkg/cloud/rgraph/exec/executor_parallel.go
bowei, line 129: toProcess => notRunnable

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+	ex.lock.Lock()
+	defer ex.lock.Unlock()
+	if runErr == nil {
+		ex.result.Completed = append(ex.result.Completed, a)
+	} else {
+		ex.result.Errors = append(ex.result.Errors, ActionWithErr{Action: a, Err: runErr})
+	}
+}
+
+func (ex *parallelExecutor) addToErrors(a Action, runErr error) {
```
bowei, line 169: this func seems to be incomplete

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+	// lock guards results
+	lock   sync.Mutex
+	result *Result
+
+	pq   *algo.ParallelQueue[Action]
+	done chan *TraceEntry
+}
+
+var _ Executor = (*parallelExecutor)(nil)
+
```
bowei, line 68: I think we should document the locking strategy used.

the parallel func will never take ex.lock

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+	queueErr := ex.pq.Run(subctx, runAction)
+	cancel()
+	if queueErr != nil {
+		waitErr := ex.pq.WaitForOrphans(ctx)
+		if waitErr != nil {
+			return ex.result, fmt.Errorf("ParallelExecutor: WaitForOrphans: %w", waitErr)
+		}
+	}
+	if len(ex.result.Errors) > 0 || len(ex.result.Pending) != 0 {
+		return ex.result, fmt.Errorf("Executor did not process all actions")
```
bowei, line 117: I think we should have a specific error that can be checked:

ErrPendingActions = errors.New("Executor did not process all actions")

so the user can check explicitly for this condition.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+		te.End = time.Now()
+		klog.V(4).Infof("Finish action %s, err: %v", a, runErr)
+
+		ex.addActionResult(a, runErr)
+
+		if runErr != nil {
+			klog.Infof("Got error  %v, from action %s error_strategy: %s", runErr, a, ex.config.ErrorStrategy)
+			// check error strategy and decide if new actions should be executed.
+			if ex.config.ErrorStrategy == StopOnError {
+				return fmt.Errorf("parallelExecutor: stopping execution for Action %s (got %v)", a, runErr)
```
bowei, line 89: Do we need to cancel the context to see if we can make the existing operations quit early?

bowei (commented):
Have we run this with the thread sanitizer?

pkg/cloud/rgraph/exec/executor.go
```diff
@@ -57,6 +59,10 @@ func DryRunOption(dryRun bool) Option {
 	return func(c *ExecutorConfig) { c.DryRun = dryRun }
 }
 
+func TimeoutOption(t time.Duration) Option {
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// lock guards results
+	lock   sync.Mutex
+	result *Result
+
+	pq   *algo.ParallelQueue[Action]
+	done chan *TraceEntry
+}
+
+var _ Executor = (*parallelExecutor)(nil)
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: removed

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: context is canceled in `(ex *parallelExecutor) Run` line 115

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: I don't understand `the parallel func will never take ex.lock`?

lock is acquired when accessing result. Result can be access by function `runAction` which is called from parallel queue so in fact this is a parallel action. But the lock is acquired for short period of time not whole operation.
There is a comment that lock is guarding the results.

kl52752:
> Have we run this with the thread sanitizer?

I run parallel_executor ut tests with go test -race options and I run I e2e test which I modify locally to use parallel_executor

go test -race -timeout 30s -run ^TestParallelExecutor* github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/exec -v
--- PASS: TestParallelExecutor (0.01s)
    --- PASS: TestParallelExecutor/empty_graph (0.00s)
    --- PASS: TestParallelExecutor/one_action (0.00s)
    --- PASS: TestParallelExecutor/action_and_dependency (0.00s)
    --- PASS: TestParallelExecutor/chain_of_3_actions (0.00s)
    --- PASS: TestParallelExecutor/two_chains_with_common_root (0.00s)
    --- PASS: TestParallelExecutor/two_node_cycle (0.00s)
    --- PASS: TestParallelExecutor/lot_of_children (0.00s)
    --- PASS: TestParallelExecutor/complex_fan_in (0.00s)
    --- PASS: TestParallelExecutor/cycle_in_larger_graph (0.00s)
    --- PASS: TestParallelExecutor/error_in_action (0.00s)
--- PASS: TestParallelExecutorErrorStrategy (0.00s)
    --- PASS: TestParallelExecutorErrorStrategy/linear_graph_StopOnError (0.00s)
    --- PASS: TestParallelExecutorErrorStrategy/linear_graph_ContinueOnError (0.00s)
    --- PASS: TestParallelExecutorErrorStrategy/branched_graph_StopOnError (0.00s)
    --- PASS: TestParallelExecutorErrorStrategy/branched_graph_ContinueOnError (0.00s)
PASS
ok      github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/exec 1.254s

go[101/328]bin/go test -race -timeout 30m -run ^TestRgraphTCPRouteAddBackends$ github.com/GoogleCloudPlatform/k8s-cloud-provider/e2e -v -project
 katarzynalach-gke-dev -resourcePrefix rgraph-tcp                                                                                    
 === RUN   TestRgraphTCPRouteAddBackends 
 --- PASS: TestRgraphTCPRouteAddBackends (212.86s)
PASS
ok      github.com/GoogleCloudPlatform/k8s-cloud-provider/e2e   214.140s

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
kl52752: In parallel executor yes. I added this in case parallel queue or runAction got stuck somehow.

aojea:
if you add this commit then the CI will do it for us anytime :)

```diff
diff --git a/Makefile b/Makefile
index b91b4ee..fa8e51a 100644
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ build: gen
 test: gen
        # Test only the library. e2e must be run in a special environment,
        # so is skipped.
-       go test ./pkg/...
+       go test -race ./pkg/...
        # We cannot use golint currently due to errors in the GCP API naming.
        # golint ./...
        go vet ./...
```

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
kl52752: I can add TODO to implement it in serial executor if needed

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

kl52752:
Thanks @aojea I will do it in separate PR

pkg/cloud/rgraph/exec/executor_parallel.go
bowei: This swallows the error -- you should chain with %w.

(ignore my comment about cancel())

bowei:
Mostly LGTM, some minor comments.

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

kl52752:
> Mostly LGTM, some minor comments.

resolved

bowei:
Please do the Makefile change in a separate PR.

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
bowei: Can this be answered?

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+		te.End = time.Now()
+		klog.V(4).Infof("Finish action %s, err: %v", a, runErr)
+
+		ex.addActionResult(a, runErr)
+
+		if runErr != nil {
+			klog.Infof("Got error  %v, from action %s error_strategy: %s", runErr, a, ex.config.ErrorStrategy)
+			// check error strategy and decide if new actions should be executed.
+			if ex.config.ErrorStrategy == StopOnError {
+				return fmt.Errorf("parallelExecutor: stopping execution for Action %s (got %w)", a, runErr)
```
bowei, line 95: * We need to be careful -- returning here measn the Tracer does not record the error.
* suggest the error be more direct: 

```
return fmt.Errorf("parallelExecutor: StopOnError due to Action %s: %w", a, runErr)
```

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if len(ex.result.Errors) > 0 || len(ex.result.Pending) != 0 {
+		return ex.result, ErrPendingActions
+	}
+	return ex.result, nil
+
+}
+
+func (ex *parallelExecutor) queueRunnableActions() {
+	klog.V(4).Infof("queueRunnableActions() pending: %d", len(ex.result.Pending))
```
bowei, line 130: Let makes this easier to understand:

klog.V(4).Infof("queueRunnableActions: %d actions pending", len...)

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+
+	ex.lock.Lock()
+	defer ex.lock.Unlock()
+
+	taskWasRun := false
+	notRunnable := []Action{}
```
bowei, line 136: var notRunnable []Action

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, a := range ex.result.Pending {
+		if a.CanRun() {
+			klog.V(4).Infof("Run task: %s", a)
+			ex.pq.Add(a)
+			taskWasRun = true
+		} else {
+			// store actions that cannot be run in a new array
```
bowei, line 143: I think this comment is not needed.

pkg/cloud/rgraph/exec/executor_parallel.go
bowei, line 130: You need to put the log line INSIDE the lock. Please fix this.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+
+		if runErr != nil {
+			klog.Infof("Got error  %v, from action %s error_strategy: %s", runErr, a, ex.config.ErrorStrategy)
+			// check error strategy and decide if new actions should be executed.
+			if ex.config.ErrorStrategy == StopOnError {
+				return fmt.Errorf("parallelExecutor: stopping execution for Action %s (got %w)", a, runErr)
+			}
+		} else {
+			// notify parents only when action finished with success
+			for _, ev := range events {
```
bowei, line 99: Should we change signal() func signature to take a list of events? Then we would only need to take the lock once vs for each Event.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	klog.V(4).Infof("Remaining pending actions: %d", len(notRunnable))
+	// update Pending array only if actions were run
+	if taskWasRun {
+		ex.result.Pending = notRunnable
+	}
+}
+
+// signal notifies parents that action finished
+func (ex *parallelExecutor) signal(ev Event) []TraceSignal {
```
bowei, line 155: suggest make this 

`signalEvents(events []Event)`

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
bowei: Please remove this for now until we can get the above question answered.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+
+func defaultParallelExecutorConfig() *ExecutorConfig {
+	return &ExecutorConfig{
+		DryRun:        false,
+		ErrorStrategy: ContinueOnError,
+		Timeout:       5 * time.Minute,
+	}
+}
+
+// NewParallelExecutor returns a new Executor that runs tasks multi`-threaded.
```
bowei, line 43: fix the extra backquote

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// parallelExecutor implements Executor.
+var _ Executor = (*parallelExecutor)(nil)
+
+func (ex *parallelExecutor) Run(ctx context.Context, c cloud.Cloud) (*Result, error) {
+	ex.queueRunnableActions()
+	subctx, cancel := context.WithTimeout(ctx, ex.config.Timeout)
+
+	runAction := func(ctx context.Context, a Action) error {
```
bowei, line 79: It would be good to just take this out and make it a method.

func (ex *parallelExecutor) runAction(ctx, a) {...}

bowei (commented):
Took a more detailed look. There are some more things to fix.

Please remove the TimeoutOption if my question cannot be answered yet.

bowei:
I would like to see some more adversarial testing. This can be a follow up.

```
a -> b
a -> c

set timeout
c blocks; b returns error

test both modes
```

Really try to break it and find race conditions.

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: I created this function here because I wanted to capture the cloud variable.
Actually I think it would be better to pass cloud in constructor "NewParallelExecutor" function and change Run to take only context. 
But this would require changes in Executor:
```
type Executor interface {
	// Run the actions. Returns non-nil if there was an error in execution of
	// one or more Actions.
	Run(context.Context) (*Result, error)
}
```

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: changed, 
Just curious, what is the reasoning behind this change? Is this purely estetic or some performance issue?

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
bowei: If I did this:

```
ctxWithTimeout := context.WithTimeout(ctx, /*my timeout*/)
ex := NewParallelExecutor()
ex.Run(ctxWithTimeout)
```

Does this do the same time as the `TimeoutOption`? When a parent is canceled -- the children also get canceled as I understand. Then we don't need a special `TimeoutOption`, instead we should document that using `context.WithTimeout` is the way people should be implementing timeouts.

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
kl52752: TimeoutOption is needed if we wants to wait for all children to finish in case of timeout. There is a function in parallel queue `WaitForOrphans` and it is to ensure that all queue goroutines have returned.  
```
// WaitForOrphans will block until remaining op() goroutines
// finish. Call this if Run() returns an error and you need to know
// that all remaining threads of execution are done.
//
// WaitForOrphans will exit early if ctx is cancelled. This `ctx`
// should be different from the `ctx` given to Run().
```

we cannot handle WaitForOrphans and Run with the same context which is provided by client becasue WaitForOrphans will immediately returns with error. I think that it is a good practice to wait for all go routines to finish or return the error that sth is still running.

kl52752:
> I would like to see some more adversarial testing. This can be a follow up.
> 
> ```
> a -> b
> a -> c
> 
> set timeout
> c blocks; b returns error
> 
> test both modes
> ```
> 
> Really try to break it and find race conditions.

hi I tested both scenarios (timeouting executor with TimeoutOptions and using context with canceled passed by user). 
Please see tests scenarios and results in this [PR](https://github.com/kl52752/k8s-cloud-provider/pull/5)

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
bowei: Ok, let's add WaitForOrphansTimeoutOption as well.

We should also document:

```
// Run executes pending actions in parallel.
//
// ParallelExecutor will stop execution (to the extent possible) if the context
// passed to Run() is cancelled. This will also cancel any waiting for orphan go
// routines that are currently executing.
```

bowei (commented):
Minor comments. Otherwise LGTM

pkg/cloud/rgraph/exec/executor.go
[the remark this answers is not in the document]
kl52752: done

kl52752:
> Minor comments. Otherwise LGTM

fixed, PTAL

pkg/cloud/rgraph/exec/executor.go
```diff
@@ -57,6 +56,19 @@ func DryRunOption(dryRun bool) Option {
 	return func(c *ExecutorConfig) { c.DryRun = dryRun }
 }
 
+// TimeoutOption sets timeout for executor Run function.
+// This option can be used with parallel executor only.
```
bowei, line 60: Can we have a follow up to make this work with the SerialExecutor 

Actually, what is actually required to make it work? Isn't this simple change?

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -0,0 +1,189 @@
...
+// parallelExecutor implements Executor.
+var _ Executor = (*parallelExecutor)(nil)
+
+// Run executes pending actions in parallel.
+//
+// ParallelExecutor will stop execution (to the extent possible) if the context
+// passed to Run() is cancelled. This will also cancel any waiting for orphan go
+// routines that are currently executing.
+//
+// To handle timeout properly use Executor Config. Set TimeoutOption for
```
bowei, line 83: ExecutorConfig is not something accessible to the user and should not be in the public docs. You should only reference the public API that you want users to use, not anything package private.

bowei:
There are still changes to be made, but let's put them in follow on PRs:

- [ ] Docstring changes (see comments I left on the PR)
- [ ] If the user does not specify a Timeout, WaitForChildren timeout, then there should be no timeout
- [ ] Make the timeout Options work (or ignored) for SerialExecutor.

PR #188, opened 2024-04-19
Move functions using global test flags to *_test file

AwesomePatrol:
Newly added helper functions were part of the package while the variables they depend on were defined in the *_test.go resulting in compile failures.

Alternative solutions:
1. Move test-helpers to helpers_test - all functions/variables in this file are unexported so it is useless as a package anyway
2. Add new arguments to newly added functions so they won't depend on global variables - helpers would be not that helpful and it would be better to inline them (as each is called only once)

/cc @kl52752
/assign @bowei

### DIFF
--- a/e2e/tcproute_test.go
+++ b/e2e/tcproute_test.go
@@ -432,3 +432,11 @@ func processGraphAndExpectActions(t *testing.T, graphBuilder *rgraph.Builder, ex
 	t.Logf("exec got Result.Errors len(%d) =\n%v", len(res.Errors), res.Errors)
 	t.Logf("exec got Result.Pending len(%d) =\n%v", len(res.Pending), res.Pending)
 }
+
+func defaultNetworkURL() string {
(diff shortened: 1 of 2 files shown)


kl52752:
thanks for fixing the build.
/lgtm

bowei:
How did we manage to commit a build break? Can we (separate PR) fix the build so this is not possible to submit?

kl52752:
> How did we manage to commit a build break? Can we (separate PR) fix the build so this is not possible to submit?

[Prow](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2117/files#diff-5a36ad3bd856fa70f17bc7c981cb133701cf01f4bd44e25c437dec51a37ff470) is doing `make test` which is not building the repo. I'm fixing that.

kl52752:
@bowei Fix [PR](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2256) PTAL

PR #195, opened 2024-04-25
fix: Add MinimumRateLimiter configured to 50ms to e2e tests

AwesomePatrol:
When running e2e tests in the CI pipeline, we often hit `RATE_LIMIT_EXCEEDED` error. The quota is 1200 requests per minute. 50ms minimum wait before making a request guarantees we won't hit it (60*1000 / 1200).

(leaking timer is not concern in tests and is fixed in go 1.23+)

Bug: b/336784945

### DIFF
--- a/e2e/testmain_test.go
+++ b/e2e/testmain_test.go
@@ -24,6 +24,7 @@ import (
 	"math/rand"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"golang.org/x/oauth2/google"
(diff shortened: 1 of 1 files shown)


e2e/testmain_test.go
```diff
@@ -77,7 +78,8 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	svc, err := cloud.NewService(ctx, client, &cloud.SingleProjectRouter{ID: testFlags.project}, &cloud.NopRateLimiter{})
+	mrl := &cloud.MinimumRateLimiter{RateLimiter: &cloud.NopRateLimiter{}, Minimum: 50 * time.Millisecond}
```
kl52752, line 81: why not pass this as a flag with default value 50ms? This value might change in future and this way we can configure it per project.

AwesomePatrol: 1. This is mitigation and not a fix. We are still investigating the root cause
2. The value matches default quota and is still low enough to not impact test speed negatively (+7.5s is smaller than +-20s variance)

PR #196, opened 2024-04-25
Parallel executor - remove default timeout value

kl52752:
Remove default timeout values for action timeout and waitForOrphans timeout. If timeout is not set then do not create context with cancel.

Run function returns Result as a pointer. Deep copy the result to avoid race conditions when waitForOrphans function timeouts and not all actions were processed. 

Add IsDone function to parallel queue to check queue status and prevent scheduling new tasks while waiting for orphans.

### DIFF
--- a/pkg/cloud/rgraph/algo/parallel_queue.go
+++ b/pkg/cloud/rgraph/algo/parallel_queue.go
@@ -108,16 +108,18 @@ const (
 	stateDone
 )
 
-// Add an item to the queue. This method is threadsafe within op() and
-// can be called during Run(). It is NOT safe to call Add() from
-// a different, unassociated thread.
+// Add an item to the queue. This method is thread safe within op() and can be
+// called during Run(). It is NOT safe to call Add() from a different,
+// unassociated thread. This method returns an error when queue state is done.
 //
-// Calling Add(item) from the op() given to Run() guarantees that item
-// will be processed.
-func (q *ParallelQueue[T]) Add(item T) {
+// Calling Add(item) from the op() given to Run() guarantees that item will be
+// processed.
+func (q *ParallelQueue[T]) Add(item T) error {
 	q.lock.Lock()
 	defer q.lock.Unlock()
-
+	if q.state == stateDone {
+		return fmt.Errorf("Queue is done")
+	}
 	qe := queueElement[T]{
 		ri: RunInfo{
 			ID:     item.String(),
@@ -136,6 +138,7 @@ func (q *ParallelQueue[T]) Add(item T) {
 		// the <-q.in will happen AFTER append(q.pending).
 		q.in <- struct{}{}
 	}
+	return nil
 }
 
 // Run the queue using op() to process each task. Different op()s must
--- a/pkg/cloud/rgraph/algo/parallel_queue.go
+++ b/pkg/cloud/rgraph/algo/parallel_queue.go
@@ -110,15 +110,15 @@ const (
 
 // Add an item to the queue. This method is thread safe within op() and can be
 // called during Run(). It is NOT safe to call Add() from a different,
-// unassociated thread. This method returns an error when queue state is done.
+// unassociated thread. This method returns False when queue state is done.
 //
 // Calling Add(item) from the op() given to Run() guarantees that item will be
 // processed.
-func (q *ParallelQueue[T]) Add(item T) error {
+func (q *ParallelQueue[T]) Add(item T) bool {
 	q.lock.Lock()
 	defer q.lock.Unlock()
 	if q.state == stateDone {
-		return fmt.Errorf("Queue is done")
+		return false
 	}
 	qe := queueElement[T]{
 		ri: RunInfo{
@@ -138,7 +138,7 @@ func (q *ParallelQueue[T]) Add(item T) error {
 		// the <-q.in will happen AFTER append(q.pending).
 		q.in <- struct{}{}
 	}
-	return nil
+	return true
 }
 
 // Run the queue using op() to process each task. Different op()s must
--- a/pkg/cloud/rgraph/exec/executor_parallel.go
+++ b/pkg/cloud/rgraph/exec/executor_parallel.go
@@ -150,7 +150,10 @@ func (ex *parallelExecutor) queueRunnableActions() {
 	for _, a := range ex.result.Pending {
 		if a.CanRun() {
 			klog.V(4).Infof("Run task: %s", a)
-			ex.pq.Add(a)
+			if err := ex.pq.Add(a); err != nil {
+				klog.Errorf("error adding task %s to parallel queue: %v", a, err)
+				break
+			}
 			taskWasRun = true
 		} else {
 			notRunnable = append(notRunnable, a)
--- a/pkg/cloud/rgraph/exec/executor_parallel.go
+++ b/pkg/cloud/rgraph/exec/executor_parallel.go
@@ -34,10 +34,8 @@ var (
 
 func defaultParallelExecutorConfig() *ExecutorConfig {
 	return &ExecutorConfig{
-		DryRun:                false,
-		ErrorStrategy:         ContinueOnError,
-		Timeout:               5 * time.Minute,
-		WaitForOrphansTimeout: 1 * time.Minute,
+		DryRun:        false,
+		ErrorStrategy: ContinueOnError,
 	}
 }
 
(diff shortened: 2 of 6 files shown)


pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,48 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
...
+		return ex.pq.Run(runCtx, ex.runAction)
+	}
+	return runWithTimeout(ctx, ex.config.Timeout, f)
+}
+
+func (ex *parallelExecutor) waitForQueueOrphans(ctx context.Context) error {
+	if ex.config.WaitForOrphansTimeout == 0 {
+		return ex.pq.WaitForOrphans(ctx)
+	}
+	klog.V(4).Infof("Run waitForQueueOrphans with timeout %v.", ex.config.WaitForOrphansTimeout)
```
AwesomePatrol, line 141: nit: `waitForOrphans` consistency?

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,48 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
...
+func (ex *parallelExecutor) runActionQueue(ctx context.Context) error {
+	if ex.config.Timeout == 0 {
+		return ex.pq.Run(ctx, ex.runAction)
+	}
+	klog.V(4).Infof("Run runAction with timeout %v.", ex.config.Timeout)
+	f := func(runCtx context.Context) error {
+		return ex.pq.Run(runCtx, ex.runAction)
+	}
+	return runWithTimeout(ctx, ex.config.Timeout, f)
+}
```
AwesomePatrol, line 135: Please consider removing `runWithTimeout` function above and trying the code below:
```suggestion
func (ex *parallelExecutor) runActionQueue(ctx context.Context) error {
	var cancel context.CancelFunc
	if ex.config.Timeout > 0 {
                ctx, cancel = context.WithTimeout(ctx, ex.config.Timeout)
		defer cancel()
		klog.V(4).Infof("Run runAction with timeout %v.", ex.config.Timeout)
	}
	return ex.pq.Run(ctx, ex.runAction)
}
```

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done - also with second function (wait for orphans)

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,42 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
...
+	}
+	copy(resultCopy.Completed, ex.result.Completed)
+	copy(resultCopy.Errors, ex.result.Errors)
+	copy(resultCopy.Pending, ex.result.Pending)
+
+	return &resultCopy
+}
+
+func (ex *parallelExecutor) runActionQueue(ctx context.Context) error {
+	var cancel context.CancelFunc
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -140,6 +172,10 @@ func (ex *parallelExecutor) runAction(ctx context.Context, a Action) error {
 }
 
 func (ex *parallelExecutor) queueRunnableActions() {
+	if ex.pq.IsDone() {
```
bowei, line 175: Is this required or an optimization?

If it's required, I'm concerned about if this is a race use to doing this outside of a lock.

kl52752: IsDone is locked by parallel_queue lock. It is thread safe and we don't have to acquire parallel_executor lock. 
This is to prevent task scheduling on Error when error strategy is StopOnError. https://screenshot.googleplex.com/9QxSKBMk46cCWdR

We need to check the queue because paralell_queue will finish (set state done) on error [link](https://github.com/GoogleCloudPlatform/k8s-cloud-provider/blob/e4456cb76f3897d74cd69320aa2f4d769c3ec1f8/pkg/cloud/rgraph/algo/parallel_queue.go#L183) and there is no other way to check that (because context is not canceled yet). If we don't check that we will try to schedule the task, but parallel_queue will not process it and we will lock the information about this task state because it will not be in pending array.

Alternative option is to return status from [func (q *ParallelQueue[T]) Add(item T) {](https://github.com/GoogleCloudPlatform/k8s-cloud-provider/blob/e4456cb76f3897d74cd69320aa2f4d769c3ec1f8/pkg/cloud/rgraph/algo/parallel_queue.go#L117-L118). This way we will know if task will be schedule for processing but I think that checking IsDone is a better way.

kl52752: I've been thinking about that, and you are right. 
I think that we should return status in paralelQueue:Add function. this way we will ne 100% sure if action will be scheduled by the queue or not.

kl52752: fixed PTAL

pkg/cloud/rgraph/algo/parallel_queue.go
```diff
@@ -108,16 +108,18 @@ const (
...
-// will be processed.
-func (q *ParallelQueue[T]) Add(item T) {
+// Calling Add(item) from the op() given to Run() guarantees that item will be
+// processed.
+func (q *ParallelQueue[T]) Add(item T) error {
 	q.lock.Lock()
 	defer q.lock.Unlock()
-
+	if q.state == stateDone {
+		return fmt.Errorf("Queue is done")
```
bowei, line 121: If this is the only kind of error that is returned, I would rather this be a bool.

`return false`

Remember to change the docstring.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,42 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
 
 }
 
+func (ex *parallelExecutor) deepCopyResult() *Result {
```
bowei, line 103: Put the copy method on Result itself (without lock)

```
func (*Result) copy() *Result {
 ret := Result{
		Completed: make([]Action, len(ex.result.Completed)),
		Pending:   make([]Action, len(ex.result.Pending)),
		Errors:    make([]ActionWithErr, len(ex.result.Errors)),
	}
	copy(resultCopy.Completed, ex.result.Completed)
	copy(resultCopy.Errors, ex.result.Errors)
	copy(resultCopy.Pending, ex.result.Pending)
  return ret
}
```

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -80,21 +78,19 @@ var _ Executor = (*parallelExecutor)(nil)
...
 	if queueErr != nil {
-		cleanUpCtx, cleanUpCancel := context.WithTimeout(ctx, ex.config.WaitForOrphansTimeout)
-		waitErr := ex.pq.WaitForOrphans(cleanUpCtx)
-		cleanUpCancel()
+		waitErr := ex.waitForQueueOrphans(ctx)
 		if waitErr != nil {
-			return ex.result, fmt.Errorf("ParallelExecutor: WaitForOrphans: %w", waitErr)
+			// Actions might still run and modify the results. Because result is
+			// returned as a pointer we need to deep copy it.
+			result := ex.deepCopyResult()
```
bowei, line 92: move the lock aquisition here.

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,42 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
...
+
+	return &resultCopy
+}
+
+func (ex *parallelExecutor) runActionQueue(ctx context.Context) error {
+	if ex.config.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, ex.config.Timeout)
+		defer cancel()
+		klog.V(4).Infof("Run runAction with timeout %v.", ex.config.Timeout)
```
bowei, line 124: you should log on both branches

pkg/cloud/rgraph/exec/executor_parallel.go
```diff
@@ -104,6 +100,42 @@ func (ex *parallelExecutor) Run(ctx context.Context) (*Result, error) {
...
+	}
+	return ex.pq.Run(ctx, ex.runAction)
+}
+
+func (ex *parallelExecutor) waitForQueueOrphans(ctx context.Context) error {
+	if ex.config.WaitForOrphansTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, ex.config.WaitForOrphansTimeout)
+		defer cancel()
+		klog.V(4).Infof("Run WaitForOrphans with timeout %v.", ex.config.WaitForOrphansTimeout)
```
bowei, line 134: same here -- log on both branches

pkg/cloud/rgraph/algo/parallel_queue.go
kl52752: I made it an error because then in place where it is used (like parallel-executor) we can print why the Add operation did not succeed. With bool we know if the operation succeeded or not but we need to know the implementation details to correlate it with the fact that the queue has finished and will not process more events. And in parallel executor we are aborting the whole operation not only this one action that did not succeed. So it is not obvious why we are doing that based on returned bool value. 

`Remember to change the docstring.`
I added the line in this method comments that it will return error when queue is done.

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: But there is only one way it can currently fail -- your error message does not give more information than the bool.

kl52752: yes, right now it can fail only due to one reason but this is an implementation detail, right? 
There are 3 states here. 
1. queue is done - nothing will be process
2. queue is running - the action will be processed 100%
3. The queue is not started - the action will be process when it starts (eventually)

Then returning the bool is not clear what it means. Will the action be processed or not? In case of error we are signaling that sth is wrong with the queue and the client should not use it, and on the other hand it is ok to return nil for other cases because the queue is in a good state.
On the other hand, returning true when the queue is in NotStarted state is not that obvious for me. I'm not clear what it means and that's why I decided to go with the error. Because the message is more clear to me.

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/exec/executor_parallel.go
kl52752: done

pkg/cloud/rgraph/algo/parallel_queue.go
bowei: I don't quite follow the reasoning here:

You can state unambiguously that Add() returns False if the Queue is done.

1. It's not an error to Add() to a queue that hasn't started.
1. By using an `error` type, the user of this API needs to treat the return as potentially either a DONE situation OR maybe some other error. But in this case, Done is the only thing that can happen.
1. Your error is static (i.e. it's just one kind of error that is being returned). It does not convey more information.

kl52752: done

bowei:
Thanks for the changes.

/lgtm
/approve

PR #199, opened 2024-04-29
Add impersonation option to e2e tests

AwesomePatrol:
It allows to impersonate a different service account to verify quota issues. See b/336784945#comment12 for more details.

Bug: b/336784945

### DIFF
--- a/e2e/testmain_test.go
+++ b/e2e/testmain_test.go
@@ -27,11 +27,14 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"k8s.io/klog/v2"
 
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
+	"google.golang.org/api/impersonate"
+	"google.golang.org/api/option"
 )
 
 var (
@@ -39,11 +42,13 @@ var (
 	theCloud cloud.Cloud
 	// testFlags passed in from the command line.
 	testFlags = struct {
-		project        string
-		resourcePrefix string
+		project            string
+		resourcePrefix     string
+		serviceAccountName string
 	}{
-		project:        "",
-		resourcePrefix: "k8scp-",
+		project:            "",
+		resourcePrefix:     "k8scp-",
+		serviceAccountName: "",
 	}
 	runID string
 )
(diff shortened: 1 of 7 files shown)


e2e/testmain_test.go
```diff
@@ -74,10 +80,25 @@ func TestMain(m *testing.M) {
...
 	ctx := context.Background()
-	client, err := google.DefaultClient(ctx, compute.ComputeScope)
+
+	credentials, err := google.FindDefaultCredentials(ctx, compute.ComputeScope)
 	if err != nil {
 		log.Fatal(err)
 	}
+	ts := credentials.TokenSource
+
+	if testFlags.serviceAccountName != "" {
```
kl52752, line 90: Can you add comment what credentials will be taken if this field is empty

e2e/testmain_test.go
```diff
@@ -74,10 +80,25 @@ func TestMain(m *testing.M) {
...
+		ts, err = impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+			TargetPrincipal: testFlags.serviceAccountName,
+			Scopes:          []string{compute.ComputeScope, compute.CloudPlatformScope},
+			Lifetime:        1200 * time.Second,
+			Delegates:       []string{},
+		}, option.WithCredentials(credentials))
+		if err != nil {
+			log.Fatal(err)
```
kl52752, line 98: please extend the log with more info + testFlags.serviceAccountName

e2e/testmain_test.go
```diff
@@ -74,10 +80,25 @@ func TestMain(m *testing.M) {
...
 	if err != nil {
 		log.Fatal(err)
 	}
+	ts := credentials.TokenSource
+
+	if testFlags.serviceAccountName != "" {
+		ts, err = impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+			TargetPrincipal: testFlags.serviceAccountName,
+			Scopes:          []string{compute.ComputeScope, compute.CloudPlatformScope},
+			Lifetime:        1200 * time.Second,
```
kl52752, line 94: 20minutes is this enough? Maybe we should parametrize this?

e2e/testmain_test.go
AwesomePatrol: the behavior stays the same as before

e2e/testmain_test.go
AwesomePatrol: it should refresh after this time elapses. I will verify tomorrow

AwesomePatrol: If `Lifetime` is not specified, it will auto-refresh (every 1 hour).

e2e/testmain_test.go
AwesomePatrol: Done (although I believe that error messages in all functions called here are already sufficiently clear)

PR #200, opened 2024-04-30
Add new types of rate limiters

AwesomePatrol:
1. `TickerRateLimiter` is a simple implementation based on time.Ticker suggested in https://go.dev/wiki/RateLimiting. It works by spacing requests in predefined intervals. It doesn't allows bursts although it is fairly easy to add. A good candidate for replacement would be https://github.com/uber-go/ratelimit which supports "slack"
2. `CompositeRateLimiter` doesn't implement any rate limiting algorithm, but allows to combine all rate limiters (implementing the `RateLimiter` interface). A user can specify what rate limiter should be used for a specific combination of project, service and operation. An example usage is available in `e2e/testmain_test.go` where there are two `TickerRateLimiter`s used for `Compute` and `NetworkServices` API; and two `MinimumRateLimiter`s for `Operations` and as a fallback for all other requests

Bug: b/337197802

### DIFF
--- a/e2e/testmain_test.go
+++ b/e2e/testmain_test.go
@@ -104,18 +104,18 @@ func TestMain(m *testing.M) {
 
 	// The default limit is 1500 per minute. Leave 200 buffer.
 	computeRL := cloud.NewTickerRateLimiter(1300, time.Minute)
-	crl.Register("", "HealthChecks", "", computeRL)
-	crl.Register("", "BackendServices", "", computeRL)
-	crl.Register("", "NetworkEndpointGroups", "", computeRL)
+	crl.Register("HealthChecks", "", computeRL)
+	crl.Register("BackendServices", "", computeRL)
+	crl.Register("NetworkEndpointGroups", "", computeRL)
 
 	// The default limit is 1200 per minute. Leave 200 buffer.
 	networkServicesRL := cloud.NewTickerRateLimiter(1000, time.Minute)
-	crl.Register("", "TcpRoutes", "", networkServicesRL)
-	crl.Register("", "Meshes", "", networkServicesRL)
+	crl.Register("TcpRoutes", "", networkServicesRL)
+	crl.Register("Meshes", "", networkServicesRL)
 
 	// To ensure minimum time between operations, wrap the network services rate limiter.
 	orl := &cloud.MinimumRateLimiter{RateLimiter: networkServicesRL, Minimum: 100 * time.Millisecond}
-	crl.Register("", "Operations", "", orl)
+	crl.Register("Operations", "", orl)
 
 	svc, err := cloud.NewService(ctx, client, &cloud.SingleProjectRouter{ID: testFlags.project}, crl)
 	if err != nil {
--- a/e2e/testmain_test.go
+++ b/e2e/testmain_test.go
@@ -98,8 +98,26 @@ func TestMain(m *testing.M) {
 		}
 	}
 	client := oauth2.NewClient(ctx, ts)
+
 	mrl := &cloud.MinimumRateLimiter{RateLimiter: &cloud.NopRateLimiter{}, Minimum: 50 * time.Millisecond}
-	svc, err := cloud.NewService(ctx, client, &cloud.SingleProjectRouter{ID: testFlags.project}, mrl)
+	crl := cloud.NewCompositeRateLimiter(mrl)
+
+	// The default limit is 1500 per minute. Leave 200 buffer.
+	computeRL := cloud.NewTickerRateLimiter(1300, time.Minute)
+	crl.Register("", "HealthChecks", "", computeRL)
+	crl.Register("", "BackendServices", "", computeRL)
+	crl.Register("", "NetworkEndpointGroups", "", computeRL)
+
+	// The default limit is 1200 per minute. Leave 200 buffer.
+	networkServicesRL := cloud.NewTickerRateLimiter(1000, time.Minute)
+	crl.Register("", "TcpRoutes", "", networkServicesRL)
+	crl.Register("", "Meshes", "", networkServicesRL)
+
+	// To ensure minimum time between operations, wrap the network services rate limiter.
+	orl := &cloud.MinimumRateLimiter{RateLimiter: networkServicesRL, Minimum: 100 * time.Millisecond}
+	crl.Register("", "Operations", "", orl)
+
+	svc, err := cloud.NewService(ctx, client, &cloud.SingleProjectRouter{ID: testFlags.project}, crl)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/pkg/cloud/ratelimit.go
+++ b/pkg/cloud/ratelimit.go
@@ -154,39 +154,35 @@ var _ RateLimiter = new(TickerRateLimiter)
 
 // CompositeRateLimiter combines rate limiters based on RateLimitKey.
 type CompositeRateLimiter struct {
-	// map[project id]map[resource name]map[operation name]RateLimiter
-	rateLimiters map[string]map[string]map[string]RateLimiter
+	// map[resource name]map[operation name]RateLimiter
+	rateLimiters map[string]map[string]RateLimiter
 	// defaultRL is used when no matching RateLimiter was found.
 	defaultRL RateLimiter
 }
 
 // NewCompositeRateLimiter creates a new CompositeRateLimiter that will use
-// provided default rate limiter if no better match is found.
+// provided default rate limiter if no better match is found. It is intended to
+// be used for a single project.
 //
 // # Example
 //
-//	defaultRL := /* default rate limiter */
-//	bsGetListRL := /* backend service rate limiter for get and list operation in project-1 */
-//	projectRL := /* rate limiter for project-1 */
-//	bsOtherProjectRL := /* rate limiter for backend service in other projects */
+//	bsDefaultRL := /* backend service default rate limiter */
+//	bsGetListRL := /* backend service rate limiter for get and list operations */
 //
 //	rl := NewCompositeRateLimiter(defaultRL)
-//	rl.Register("project-1", "", "", projectRL)
-//	rl.Register("project-1", "BackendServices", "Get", bsGetListRL)
-//	rl.Register("project-1", "BackendServices", "List", bsGetListRL)
-//	rl.Register("", "BackendServices", "", bsOtherProjectRL)
+//	rl.Register("BackendServices", "", bsDefaultRL)
+//	rl.Register("BackendServices", "Get", bsGetListRL)
+//	rl.Register("BackendServices", "List", bsGetListRL)
 //
 // This rate limiter is not nesting. Only one rate limiter is used for any
-// particular combination of: project, resource, operation. For the case above,
-// rate limiter registered at ("project-1", "", "") won't be applied to
-// operation ("project-1", "BackendServices", "Get"), because a more specific
-// rate limiter was registered.
+// particular combination of: resource, operation. For the case above, rate
+// limiter registered at ("BackendServices", "") won't be applied to operation
+// ("BackendServices", "Get"), because a more specific rate limiter was
+// registered.
 func NewCompositeRateLimiter(defaultRL RateLimiter) *CompositeRateLimiter {
-	m := map[string]map[string]map[string]RateLimiter{
+	m := map[string]map[string]RateLimiter{
 		"": {
-			"": {
-				"": defaultRL,
-			},
+			"": defaultRL,
 		},
 	}
 	return &CompositeRateLimiter{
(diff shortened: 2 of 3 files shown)


pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,108 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+func (t *TickerRateLimiter) Accept(ctx context.Context, rlk *RateLimitKey) error {
+	select {
+	case <-t.ticker.C:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// Observe does nothing.
```
kl52752, line 136: // Observe implements RateLimiter

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,108 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+func (*TickerRateLimiter) Observe(context.Context, error, *RateLimitKey) {
+}
+
+// CompositeRateLimiter combines rate limiters based on RateLimitKey.
+type CompositeRateLimiter struct {
+	rateLimiters map[string]map[string]map[string]RateLimiter
```
kl52752, line 142: please add in comment what those string stands for. I understand that the order is project, service and operation? It doesn't align with RateLimitKey variables order and is not clear when looking at this.

pkg/cloud/ratelimit_test.go
```diff
@@ -82,3 +82,143 @@ func TestMinimumRateLimiter(t *testing.T) {
...
+	cancelled()
+	// Verify context is cancelled by now.
+	<-ctxCancelled.Done()
+	err := trl.Accept(ctxCancelled, nil)
+	if err != ctxCancelled.Err() {
+		t.Errorf("AcceptRateLimiter.Accept() = %v, want %v", err, ctxCancelled.Err())
+	}
+}
+
+func TestCompositeRateLimiter(t *testing.T) {
```
kl52752, line 116: I think this should be parameter test where you register rl with different RL key.
you need to combine all keys in map (project, service, operation) and check that correct rl was chosen.

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,108 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+}
+
+func (c *CompositeRateLimiter) fillMissing() {
+	for _, subProject := range c.rateLimiters {
+		if subProject[""] == nil {
+			subProject[""] = map[string]RateLimiter{}
+		}
+		for _, subService := range subProject {
+			if subService[""] == nil {
+				subService[""] = c.defaultRL
```
kl52752, line 178: what is expected bahaviour here? defaultURL is a pointer so you copy pointer to all subServices. I would expect that you copy/create a new instance of RL so you can keep track request per Service and I don't think that this is working as expected. 
If you don't want separate RL instance for Service then this logic is too complex and you don't need to fill missing fields.

kl52752: Please add complex test which will cover this case.

pkg/cloud/ratelimit_test.go
```diff
@@ -82,3 +82,143 @@ func TestMinimumRateLimiter(t *testing.T) {
...
+	}
+	if calledB {
+		t.Errorf("`calledB` = true, want false")
+	}
+
+	calledA = false
+	calledC := false
+	fc := &FakeAcceptor{accept: func() { calledC = true }}
+	crl := &AcceptRateLimiter{fc}
+	rl.Register("", "networks", "", crl)
```
kl52752, line 168: you need to test that rate limiters for service and independent and does not impact on each other.

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,27 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
 func (m *MinimumRateLimiter) Observe(ctx context.Context, err error, key *RateLimitKey) {
 	m.RateLimiter.Observe(ctx, err, key)
 }
+
+type TickerRateLimiter struct {
```
bowei, line 116: Docstring describing what this does.

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,27 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+	ticker *time.Ticker
+}
+
+func NewTickerRateLimiter(limit int, interval time.Duration) *TickerRateLimiter {
```
bowei, line 120: Same here -- you need to document in a comment what this does.

pkg/cloud/ratelimit_test.go
```diff
@@ -82,3 +82,33 @@ func TestMinimumRateLimiter(t *testing.T) {
...
+
+	trl := NewTickerRateLimiter(100, time.Second)
+	start := time.Now()
+	for i := 0; i < 50; i++ {
+		err := trl.Accept(context.Background(), nil)
+		if err != nil {
+			t.Errorf("TickerRateLimiter.Accept = %v, want nil", err)
+		}
+	}
+	elapsed := time.Since(start)
```
bowei, line 97: It would be good to avoid timing-based tests as much as possible. What is the timing skew margin that this is sensitive to? We need to make it as robust to skew as possible.

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,108 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+func (*TickerRateLimiter) Observe(context.Context, error, *RateLimitKey) {
+}
+
+// CompositeRateLimiter combines rate limiters based on RateLimitKey.
+type CompositeRateLimiter struct {
+	rateLimiters map[string]map[string]map[string]RateLimiter
+	defaultRL    RateLimiter
+}
+
+// NewCompositeRateLimiter creates a new CompositeRateLimiter that will use
```
bowei, line 146: It helps to sketch out how this is used in the docstring:

```
// Example
//
// defaultRL := /*default rate limiter*/
// bsRL := /* backend service rate limiter */
//
// rl := NewCompositeRateLimiter(defaultRL)
// rl.Register("project-1", "BackendServices", "Get", bsRL)
```

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +112,108 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+		}
+	}
+}
+
+// Register adds provided rl to the composite rate limiter.
+//
+// It replaces previous rate limiter provided for the same project, service,
+// operation combination. Once a rate limiter is added, it can't be removed.
+//
+// Same rate limiter can be used for multiple Register calls.
```
bowei, line 189: It seems like there is a special semantic on matching when you pass in an empty string "". Can we document this contract in the comments?

pkg/cloud/ratelimit_test.go
AwesomePatrol: That's what I do in `TestCompositeRateLimiter_Table` below

pkg/cloud/ratelimit_test.go
AwesomePatrol: It should take a minimum of 500ms, so I check for the range [500ms, 1s]. I couldn't break. I think that mocking time or timer would make it unnecessarily complex while we only want to verify that `NewTickerRateLimiter` correctly converts its arguments to ticker period.

pkg/cloud/ratelimit.go
AwesomePatrol: "keep track request per Service" - I don't understand. The goal is specifically to share rate limiters. Otherwise there will be a lot of issues with leaking resources (and growing resource use) for each path added.

"this logic is too complex" - the main goal of this function is to avoid complexity. By prefilling entries for empty keys we make `Accept` function significantly simpler

pkg/cloud/ratelimit.go
AwesomePatrol: I believe that `Observe does nothing` is a better comment as it doesn't implement anything (only satisfies an interface - I added separate check for that).

pkg/cloud/ratelimit_test.go
AwesomePatrol: Added missing checks

pkg/cloud/ratelimit.go
kl52752: My point is that you are `ensuring exists` and then `filling missing` where you just put default rate limiter to the map and later in n`Accept` function you check for each key in map and set default as empty string to get the RateLimiter.
I would expect that if you ensure that all rate limiters are in the map you don't do the check in Accept or just don't do the `ensure` and `filling` and return default RateLimitter if key does not exist.

This is the Accept function I'm referring to (added some comments inline): 
```
func (c *CompositeRateLimiter) Accept(ctx context.Context, rlk *RateLimitKey) error {
	if rlk == nil {
		return c.defaultRL.Accept(ctx, rlk)
	}
	project := rlk.ProjectID
	if _, ok := c.rateLimiters[project]; !ok {
		project = ""  // You are calling fillMissing function but still there is a need to change project to default?
	}
	service := rlk.Service
	if _, ok := c.rateLimiters[project][service]; !ok {
		service = ""
	}
	operation := rlk.Operation
	if _, ok := c.rateLimiters[project][service][operation]; !ok {
		operation = "" // you don't ensure/fill missing operation as project + service
	}
	return c.rateLimiters[project][service][operation].Accept(ctx, rlk)
}
```

So either simplify Accept function because you rely on Register function and verify rl settings like project once at the beginning or return default whenb no project is defined.

pkg/cloud/ratelimit.go
```diff
@@ -112,3 +114,148 @@ func (m *MinimumRateLimiter) Accept(ctx context.Context, key *RateLimitKey) erro
...
+//	defaultRL := /* default rate limiter */
+//	bsGetListRL := /* backend service rate limiter for get and list operation in project-1 */
+//	projectRL := /* rate limiter for project-1 */
+//	bsOtherProjectRL := /* rate limiter for backend service in other projects */
+//
+//	rl := NewCompositeRateLimiter(defaultRL)
+//	rl.Register("project-1", "", "", projectRL)
+//	rl.Register("project-1", "BackendServices", "Get", bsGetListRL)
+//	rl.Register("project-1", "BackendServices", "List", bsGetListRL)
+//	rl.Register("", "BackendServices", "", bsOtherProjectRL)
```
kl52752, line 177: From business point of view is there a need to share one rate limiter across project? I can see in the example that you are not sharing but why you accept empty project name? This implies that we can share RL across projects and IMO it's not correct.

pkg/cloud/ratelimit.go
kl52752: why `resource name` when you call this service in code. I think service is a good choice

pkg/cloud/ratelimit.go
kl52752: we can see that this function does nothing because it is empty what we don't know is why this function is empty and I think that it is good to put this in comment otherwise the comment is redundant.

e2e/testmain_test.go
```diff
@@ -98,8 +98,26 @@ func TestMain(m *testing.M) {
...
 	}
 	client := oauth2.NewClient(ctx, ts)
+
 	mrl := &cloud.MinimumRateLimiter{RateLimiter: &cloud.NopRateLimiter{}, Minimum: 50 * time.Millisecond}
-	svc, err := cloud.NewService(ctx, client, &cloud.SingleProjectRouter{ID: testFlags.project}, mrl)
+	crl := cloud.NewCompositeRateLimiter(mrl)
+
+	// The default limit is 1500 per minute. Leave 200 buffer.
+	computeRL := cloud.NewTickerRateLimiter(1300, time.Minute)
+	crl.Register("", "HealthChecks", "", computeRL)
```
kl52752, line 107: Why don't you pass project here?

AwesomePatrol: I want it to work for all projects, not only a specific one. In this case I could use `testFlags.project`, but I also wanted this piece of code to serve as an example

pkg/cloud/ratelimit.go
AwesomePatrol: > From business point of view is there a need to share one rate limiter across project?

Yes, quotas are very often defined for a set of resources and/or operations.

> why you accept empty project name?

Providing an empty project name is a way to set a "default" if there is no more specific rate limiter provided. This is explained in the other docstring comment.

pkg/cloud/ratelimit.go
AwesomePatrol: > You are calling fillMissing function but still there is a need to change project to default?

Yes, that's the point.

> you don't ensure/fill missing operation as project + service

I do. In exactly the line that this comment is pinned to.

> I would expect that if you ensure that all rate limiters are in the map you don't do the check in Accept

Please provide a more comprehensive code snippet that would satisfy the following requirements:
1. Setting default rate limiter for a project
2. Setting default rate limiter for a service
3. Setting default rate limiter for an operation
4. Setting default rate limiter for service+operation combination

pkg/cloud/ratelimit.go
kl52752: I check public docs and from what I understand all quotas are defined per project https://screenshot.googleplex.com/7dN4SBk2oamcguC. So you should not share RateLimiter across projects. You should set default RL in context of projects. That means that every project should have independent instance of default RL. Projects passed to client are well defined so this code doesn't have to be super generic and handle all cases. Using this RL should be correlated with ProjectRouter because if you want to support multiple project you should support projects that you will use. (and in fact in most cases we are using Single project router or project router when we have network defined in different project)

pkg/cloud/ratelimit.go
AwesomePatrol: ok, that sounds reasonable. There are two options I see:
1. remove project from the path specification (rate limiters will be shared across all projects)
2. refactor this to use "rate limiter factory" instead of a rate limiter instances (this would require allocating a lot of stuff for the first call for each new project)

AwesomePatrol: Implemented 1 in this PR. I started working on 2 in https://github.com/AwesomePatrol/k8s-cloud-provider/compare/add-new-rate-limiter-types...AwesomePatrol:k8s-cloud-provider:add-per-project-rate-limiter?expand=1

e2e/testmain_test.go
```diff
@@ -98,8 +98,26 @@ func TestMain(m *testing.M) {
...
+	crl.Register("NetworkEndpointGroups", "", computeRL)
+
+	// The default limit is 1200 per minute. Leave 200 buffer.
+	networkServicesRL := cloud.NewTickerRateLimiter(1000, time.Minute)
+	crl.Register("TcpRoutes", "", networkServicesRL)
+	crl.Register("Meshes", "", networkServicesRL)
+
+	// To ensure minimum time between operations, wrap the network services rate limiter.
+	orl := &cloud.MinimumRateLimiter{RateLimiter: networkServicesRL, Minimum: 100 * time.Millisecond}
+	crl.Register("Operations", "", orl)
```
kl52752, line 118: is this intended that resource_name = Operations and it's not operations_name?

AwesomePatrol: yes, that's correct. `Operations` is the name of resource: https://cloud.google.com/resource-manager/reference/rest/v1/operations

PR #202, opened 2024-05-07
Adds OWNERS to subcomponents

bowei:


vendor/OWNERS
```diff
@@ -0,0 +1,8 @@
+approvers:
+- kl52752
+- briantkennedy
+- mag-kol
```
AwesomePatrol, line 4: 1. another approver would still be needed for `go.mod`, `go.sum` changes
2. please consider sorting in lexicographical order, so it will be easier to compare the lists

AwesomePatrol (approved):

PR #203, opened 2024-05-08
Fix race condition in parallel tests.

kl52752:
Parallel executor tests checks that when timeout occurs parallel executor will exit even if long lasting operation scheduled in separate go routine is running. This go routine might finish after the whole test is finished and this is expected behaviour. In this go routine we should not use t.Log which will be invoked at the end of the action because the test might already exit and this can cause panic. https://screenshot.googleplex.com/3tRWo69XkSfvacf

### DIFF
--- a/pkg/cloud/rgraph/exec/executor_parallel_test.go
+++ b/pkg/cloud/rgraph/exec/executor_parallel_test.go
@@ -19,6 +19,7 @@ package exec
 import (
 	"context"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
@@ -156,6 +157,32 @@ func TestParallelExecutorErrorStrategy(t *testing.T) {
 	}
 }
 
+type syncTestActions struct {
+	lock   sync.Mutex
+	active int
+	c      chan int
+}
+
+func (s *syncTestActions) launch() {
+	s.lock.Lock()
+	s.active++
+	s.lock.Unlock()
+	s.c <- 1
+}
+func (s *syncTestActions) finish() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if s.active > 0 {
+		s.active--
+	}
+}
+
+func (s *syncTestActions) accept() {
+	go func() {
+		<-s.c
+	}()
+}
+
 func TestParallelExecutorTimeoutOptions(t *testing.T) {
 	for _, tc := range []struct {
 		name string
@@ -164,6 +191,9 @@ func TestParallelExecutorTimeoutOptions(t *testing.T) {
 		waitForOrphansTimeout time.Duration
 		injectError           bool
 		wantErr               bool
+
+		finishBeforeTimeout bool
+		finishBeforeOrphans bool
 		// actions should be sorted alphabetically for comparison.
 		completed []string
 		pending   []string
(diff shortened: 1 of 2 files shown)


AwesomePatrol (approved):

bowei:
Did you upload the whole PR? There is only one change (delete Log line)

kl52752:
> Did you upload the whole PR? There is only one change (delete Log line)

yes, the problem is that this one log line in some test is called after test return and we believe that this is causing panic. https://screenshot.googleplex.com/3tRWo69XkSfvacf. Calling t.Log before sleep is safe because we know that this line will always be called during test execution. But sleep is not interrupted by context cancel and when in test we expect that executor will exit after timeout which is less than 5sec we abandon the go routine and it can finish after whole test exits. This race happens once in prow we didn't reproduced it locally.

bowei:
The construction of the test is a bit concerning as it depends very much on timing. Is there a way to redo as much of it as possible in terms of  happens before relationships?

something like this:

```
package exec

import (
	"fmt"
	"sync"
)

func NewSync() *Sync { return &Sync{states: map[string]*state{}} }

type Sync struct {
	lock   sync.Mutex
	states map[string]*state
}

type state struct {
	done    bool
	waiters []*waiter
}

type waiter struct {
	ch chan struct{}
}

// Waiters

func (s *Sync) Wait(key string) {
	s.lock.Lock()

	st, ok := s.states[key]
	if ok && st.done {
		s.lock.Unlock()
		return
	}

	if !ok {
		st := &state{}
		s.states[key] = st
	}

	// Add to the waiters and block.
	w := &waiter{make(chan struct{})}
	st.waiters = append(st.waiters, w)

	s.lock.Unlock()
	<-w.ch
}

func (s *Sync) Signal(key string) {
	s.lock.Lock()

	st, ok := s.states[key]
	if !ok {
		s.states[key] = &state{done: true}
		s.lock.Unlock()
		return
	}

	st.done = true
	for _, w := range st.waiters {
		close(w.ch)
	}
	s.lock.Unlock()
}

func example() {
	s := NewSync()

	go func() {
		s.Wait("EVENT")
		fmt.Println("Happens after EVENT")
	}()
	go func() {
		fmt.Println("EVENT signaled")
		s.Signal("EVENT")
	}()
}
```

kl52752:
@bowei this test checks how executor behaves when timeout occurs so by definition this test depends on timing. 
I think that your proposal is over complicated for this simple test (we have only 1 action that needs to be time consuming) and it does not remove time dependency completely because still we need to control when task will be finished. I don't think that spinning another go routines with channel sin test is better solutions than using sleep here.

I added struct that checks that this (one) go routine has finished before test exits so we should not hit the issue when we trie to log after test exits.

bowei:
(I'm recommending the above code as a utility that can be used in more than one test.)

kl52752:
@bowei I fix the test with reccomended approach, PTAL

pkg/cloud/rgraph/exec/executor_parallel_test.go
```diff
@@ -156,6 +157,32 @@ func TestParallelExecutorErrorStrategy(t *testing.T) {
 	}
 }
 
+type syncTestActions struct {
```
bowei, line 160: You should document this -- it's not obvious why this test object behaves like this?

bowei:
Please take a look at the comments...

kl52752:
deprecated with https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/209

PR #204, opened 2024-05-08
 Change type traits constrains logic.

kl52752:
Loosen up default type traits constraint logic that requires the field to be set.

This PR has multiple changes:
1. Introducing new type trait FieldTypeMandatory.
2. Change default type FieldTypeOrdinary to allow zero value.
After changes in type traits code coverage for health check and backend service fell below 80% so in 2 separate commits tests for given components were added.

### DIFF
--- a/pkg/cloud/api/fill.go
+++ b/pkg/cloud/api/fill.go
@@ -179,7 +179,7 @@ func fillNullAndForceSend(traits *FieldTraits, v reflect.Value) error {
 			fType := traits.fieldType(p.Field(ft.Name))
 			fv := v.Field(i)
 
-			if fType == FieldTypeOrdinary {
+			if fType == FieldTypeNonZeroValue {
 				switch {
 				case fv.IsZero() && fv.Type().Kind() == reflect.Pointer:
 					nullFields[ft.Name] = true
--- a/pkg/cloud/api/type_trait.go
+++ b/pkg/cloud/api/type_trait.go
@@ -138,8 +138,8 @@ type fieldTrait struct {
 type FieldType string
 
 const (
-	// FieldTypeOrdinary is a ordinary field. It must be set (non-zero or in a
-	// meta-field. It will be compared by value in a diff.
+	// FieldTypeOrdinary is a ordinary field. It will be compared by value in a
+	// diff. It can be zero-value without being in a metafield.
 	FieldTypeOrdinary FieldType = "Ordinary"
 	// FieldTypeSystem fields are internal infrastructure related fields. These are never
 	// copied or diff'd.
@@ -148,8 +148,13 @@ const (
 	// should never be set by the client.
 	FieldTypeOutputOnly FieldType = "OutputOnly"
 	// FieldTypeAllowZeroValue is an ordinary field that can be zero-value
-	// without being in a metafield. This is used for testing.
+	// without being in a metafield. This is used for testing. TODO(kl52752)
+	// remove this field when all resources are migrating to
+	// FieldTypeNonZeroValue.
 	FieldTypeAllowZeroValue FieldType = "AllowZeroValue"
+	// FieldTypeNonZeroValue is a field that's value must be non-zero or
+	// specified in a meta-field. It will be compared by value in a diff.
+	FieldTypeNonZeroValue FieldType = "NonZeroValue"
 )
 
 // CheckSchema validates that the traits are valid and match the schema of the
@@ -179,6 +184,9 @@ func (dt *FieldTraits) System(p Path) { dt.add(p, FieldTypeSystem) }
 
 // AllowZeroValue specifies the type of the given path.
 func (dt *FieldTraits) AllowZeroValue(p Path) { dt.add(p, FieldTypeAllowZeroValue) }
+
+// NonZeroValue specifies the type of the given path.
+func (dt *FieldTraits) NonZeroValue(p Path) { dt.add(p, FieldTypeNonZeroValue) }
 
 // Clone create an exact copy of the traits.
 func (dt *FieldTraits) Clone() *FieldTraits {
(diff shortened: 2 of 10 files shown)


pkg/cloud/api/type_trait.go
```diff
@@ -148,8 +148,12 @@ const (
 	// should never be set by the client.
 	FieldTypeOutputOnly FieldType = "OutputOnly"
 	// FieldTypeAllowZeroValue is an ordinary field that can be zero-value
```
mag-kol, line 150: This does not make much sense now, as currently ordinary field can be a zero-value.

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
```diff
@@ -43,136 +43,30 @@ func (*typeTrait) FieldTraits(v meta.Version) *api.FieldTraits {
 	dt.OutputOnly(api.Path{}.Pointer().Field("SecurityPolicy"))
 	dt.OutputOnly(api.Path{}.Pointer().Field("SelfLink"))
 
-	dt.AllowZeroValue(api.Path{}.Pointer().Field("Iap"))
```
mag-kol, line 46: I want to make sure I understand that removals in this file - we are removing the AllowZeroValue because now the FieldTypeOrdinary allows zero values?

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
```diff
@@ -43,136 +43,30 @@ func (*typeTrait) FieldTraits(v meta.Version) *api.FieldTraits {
...
-	dt.AllowZeroValue(api.Path{}.Pointer().Field("SecuritySettings").Pointer().Field("SubjectAltNames"))
-
-	if v == meta.VersionBeta || v == meta.VersionAlpha {
-		dt.AllowZeroValue(api.Path{}.Pointer().Field("SecuritySettings").Pointer().Field("Authentication"))
-		dt.AllowZeroValue(api.Path{}.Pointer().Field("CircuitBreakers").Pointer().Field("ConnectTimeout"))
-		dt.AllowZeroValue(api.Path{}.Pointer().Field("CircuitBreakers").Pointer().Field("ConnectTimeout").Pointer().Field("Nanos"))
-		dt.AllowZeroValue(api.Path{}.Pointer().Field("CircuitBreakers").Pointer().Field("ConnectTimeout").Pointer().Field("Seconds"))
-		dt.AllowZeroValue(api.Path{}.Pointer().Field("Subsetting").Pointer().Field("SubsetSize"))
-
+	dt.Mandatory(api.Path{}.Pointer().Field("LoadBalancingScheme"))
```
mag-kol, line 50: The mandatory fields are only those fields which are mandatory in reality in the GCE API?

mag-kol (approved):
Thanks! Looks good- main main comments are mostly about making sure I understand the code correctly.

pkg/cloud/api/type_trait.go
kl52752: I added TODO (line below) to remove this type when we migrate all resources. I just updated BackendService and HealthCheck in this PR. Remaining type I will do in next.

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
kl52752: yes now FieldTypeOrdinary behaves as FieldTypeAllowZeroValue.
That mean that when API changed and a new field will be added UT tests will not be failing because we will trat new fields as AllowZeroValue.
As mentioned in comment before I will remove AllowZeroValue type when all resources will be migrated.

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
kl52752: for BackendService I made fields mandatory when they are mandatory or have non zero default value (like Protocol and CompressionMode)

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
mag-kol: sgtm. thanks for claryfying

pkg/cloud/rgraph/rnode/backendservice/type_trait.go
mag-kol: SG. we clarified it during VC.

mag-kol (approved):

pkg/cloud/api/fill.go
```diff
@@ -179,7 +179,7 @@ func fillNullAndForceSend(traits *FieldTraits, v reflect.Value) error {
 			fType := traits.fieldType(p.Field(ft.Name))
 			fv := v.Field(i)
 
-			if fType == FieldTypeOrdinary {
+			if fType == FieldTypeMandatory {
```
bowei, line 182: Fill's contract is to fill __all__ field values, not just Mandatory ones. If this is not clear, please update the Docstring.

pkg/cloud/api/type_trait.go
```diff
@@ -148,8 +148,12 @@ const (
...
-	// without being in a metafield. This is used for testing.
+	// without being in a metafield. This is used for testing. TODO(kl52752)
+	// remove this field when all resources are migrating to FieldTypeMandatory.
 	FieldTypeAllowZeroValue FieldType = "AllowZeroValue"
+	// FieldTypeMandatory is a field that's value must be non-zero or specified
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/api/fill.go
kl52752: I changed the name from Mandatory to NonZeroValue so maybe this way it will be clear.
In this check we are adding fields to NullFields or ForceSendFields if they are nil. We don't have to do this for fields that are allowed to be zero

bowei:
Is TypeOrdinary the same as TypeAllowZeroValue?

Do we need both if they are the same?

kl52752:
> Is TypeOrdinary the same as TypeAllowZeroValue?
> 
> Do we need both if they are the same?

the idea was to leave field type ordinary and get rid of AllowZeroValue when we migrate all nodes to new type.

bowei:
/lgtm
/approve

It seems like there is a significant cleanup that needs to follow this.

PR #205, opened 2024-05-16
Implement retriable action

kl52752:
RetriableAction is a wrapper on Action which contains RetryProvider to decide if the action should be rerun on error. Retiable Action works with all already supported actions: Create, Delete and Update.

### DIFF
--- a/pkg/cloud/rgraph/exec/retrieable_action.go
+++ b/pkg/cloud/rgraph/exec/retrieable_action.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2024 Google LLC
+
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package exec
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+)
+
+// retriableAction is an action with retry mechanism
+type retriableAction struct {
+	Action
+	canRetry func(error) (bool, time.Duration)
+}
+
+// NewRetriableAction is an Action which check if a given action can be retired
+// after error. On error the action will be retried when canRetry(err) returns
+// true and duration for backoff. Duration equals 0 means that the action needs
+// to be retried right away.
+func NewRetriableAction(a Action, canRetry func(error) (bool, time.Duration)) Action {
+	return &retriableAction{a, canRetry}
+}
+
+// Run executes Action. On error `canRetry` function is used to check time
+// period after which the action should be retried. If canRetry returns false or
+// context is canceled action returns with error.
+func (ra *retriableAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	for {
+		events, err := ra.Action.Run(ctx, c)
+
+		if err == nil {
+			return events, nil
+		}
+		if canRetry, backOffTime := ra.canRetry(err); canRetry {
+			timer := time.NewTimer(backOffTime)
+			select {
+			case <-timer.C:
+				timer.Stop()
+				continue
+			case <-ctx.Done():
+				timer.Stop()
+				return nil, fmt.Errorf("context canceled")
+			}
+		}
+		return events, err
+	}
+}
+
+// String wraps Action name with retry information
+func (ra *retriableAction) String() string {
+	return ra.Action.String() + " with retry"
+}
--- a/pkg/cloud/rgraph/exec/retrieable_action_test.go
+++ b/pkg/cloud/rgraph/exec/retrieable_action_test.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2024 Google LLC
+
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package exec
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+)
+
+// fakeAction will return error for n actions defined in errorRunThreshold,
+// runCtr counts all action executions.
+// errorRunThreshold set to -1 means that Action should always return error.
+type fakeAction struct {
+	runCtr            int
+	errorRunThreshold int
+}
+
+func (fa *fakeAction) CanRun() bool {
+	return true
+}
+
+func (fa *fakeAction) Signal(e Event) bool {
+	return false
+}
+
+func (fa *fakeAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	fa.runCtr++
+	if fa.errorRunThreshold > fa.runCtr {
+		return EventList{}, fmt.Errorf("Action in error")
+	}
+	return EventList{}, nil
+}
+
+func (fa *fakeAction) DryRun() EventList {
+	return EventList{}
+}
+
+func (fa *fakeAction) String() string {
+	return "fakeAction"
+}
+
+func (fa *fakeAction) PendingEvents() EventList {
+	return EventList{}
+}
+
+func (fa *fakeAction) Metadata() *ActionMetadata {
+	return &ActionMetadata{
+		Name: "fakeAction",
+	}
+}
+
+// fakeRetryProvider provides retry mechanism for tests.
+// ctr - counts all retry provider calls
+// shouldRetry - tells if Action should be rerun
+type fakeRetryProvider struct {
+	ctr         int
+	shouldRetry bool
+	backOff     time.Duration
+}
+
+// IsRetriable returns info if action should be rerun. Every call to this
+// function increments counter.
+func (frp *fakeRetryProvider) IsRetriable(error) (bool, time.Duration) {
+	frp.ctr++
+	return frp.shouldRetry, frp.backOff
+}
+
+func TestRetriableAction(t *testing.T) {
+	for _, tc := range []struct {
+		name               string
+		shouldRetry        bool
+		wantError          bool
+		wantRunThreshold   int
+		wantRetriableCalls int
+		wantRun            int
+	}{
+		{
+
+			name:               "should not retry",
+			shouldRetry:        false,
+			wantError:          true,
+			wantRunThreshold:   5,
+			wantRetriableCalls: 1,
+			wantRun:            1,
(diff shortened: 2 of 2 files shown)


pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,86 @@
...
+type RetryProvider interface {
+	IsRetriable(error) bool
+}
+
+// retriableAction implements Action
+var _ Action = (*retriableAction)(nil)
+
+// retriableAction is a decorator to the action adding retry mechanism
+type retriableAction struct {
+	a             Action
```
AwesomePatrol, line 33: Please consider [embedding](https://go.dev/doc/effective_go#embedding) instead: (example: https://github.com/GoogleCloudPlatform/k8s-cloud-provider/commit/f477355b14093099abff558d764c7df1a2b3e5a0)

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,86 @@
...
+func (ra *retriableAction) Signal(e Event) bool {
+	return ra.a.Signal(e)
+}
+
+// Run executes Action. If error is returned retry provider checks if Action
+// should be rerun
+func (ra *retriableAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	var err error
+	var events EventList
+	for run := true; run; run = ra.retryProvider.IsRetriable(err) && ctx.Err() == nil {
```
AwesomePatrol, line 58: Do we plan to introduce backoff mechanism here or in the underlying client library (`pkg/cloud`)?

pkg/cloud/rgraph/exec/retrieable_action.go
kl52752: done

pkg/cloud/rgraph/exec/retrieable_action_test.go
```diff
@@ -0,0 +1,151 @@
...
+		wantRetries      int
+		wantRun          int
+	}{
+		{
+
+			name:             "should not retry",
+			shouldRetry:      false,
+			wantError:        true,
+			wantRunThreshold: 5,
+			wantRetries:      1,
```
AwesomePatrol, line 98: nit: this is a little confusing, I would expect `0` here (because there is `4` below). Maybe counting only "retry = true" or changing to `wantRetriableCalls` would help?

kl52752: I will go with `wantRetriableCalls` becaue you need at least 1 call to check if retry is needed.

pkg/cloud/rgraph/exec/retrieable_action_test.go
```diff
@@ -0,0 +1,151 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
+	_, err := ra.Run(ctx, nil)
+	cancel()
+
+	if err != nil {
+		t.Fatalf("ra.Run(context.Background(), nil) = %v, want nil", err)
+	}
+
+	if fa.runCtr > 1 {
+		t.Errorf("action run mismatch: got %v, want 4", fa.runCtr)
```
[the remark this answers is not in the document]
kl52752: done

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,57 @@
...
+func NewRetriableAction(a Action, rp RetryProvider) Action {
+	return &retriableAction{a, rp}
+}
+
+// Run executes Action. If error is returned retry provider checks if Action
+// should be rerun
+func (ra *retriableAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	var err error
+	var events EventList
+	for run := true; run; run = ra.retryProvider.IsRetriable(err) && ctx.Err() == nil {
```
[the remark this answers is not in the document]
kl52752: yes, good idea

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,56 @@
...
+package exec
+
+import (
+	"context"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+)
+
+// RetryProvider decides if action should be retired on error
+type RetryProvider interface {
```
bowei, line 24: Suggestion for an alternative since the interface is just a func:

```
NewRetryAction(a Action, canRetry func(error) bool)
```

bowei: You don't need the extra interface in this case...

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,51 @@
...
+// NewRetriableAction decorates Action with retry provider function
+func NewRetriableAction(a Action, rp func(error) bool) Action {
+	return &retriableAction{a, rp}
+}
+
+// Run executes Action. If error is returned retry provider checks if Action
+// should be rerun
+func (ra *retriableAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	for {
+		events, err := ra.Action.Run(ctx, c)
```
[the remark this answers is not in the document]
kl52752: retry policy is implemented by rate limiter. Every time you call GCE API it is going through Rate limiter.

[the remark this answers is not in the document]
kl52752: if needed it can be also implemented in retry provider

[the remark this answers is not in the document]
mag-kol: Retry Provider is an simple function as of now.

Can you point me to the code where retry policy is implemented within rate limiter? 

Do we retry actions in rate limiter when we are getting quota errors (I also couldn't find it).

Can't we have retry policy in both retriable action and rate limiter? Sometimes from rate limiter perspective we do not have to wait longer (as we still have capacity to call GCE API), but from an action perspective it maybe have some sense to wait before retrying.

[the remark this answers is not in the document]
kl52752: good point, retry action which will be used is not implemented yet. We can add there back off logic

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+)
+
+// retriableAction is a decorator to the action adding retry mechanism
+type retriableAction struct {
+	Action
+	isRetriable func(error) bool
+}
+
+// NewRetriableAction decorates Action with retry provider function
```
bowei, line 29: This comment docstring doesn't describe what the `rp` parameter means. 

Please always write the public docstrings in a way that someone reading the godoc page will be able to use the package without reading the actual underlying implementation. It also serves as a guide for defining the semantics of function.

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,51 @@
...
+func NewRetriableAction(a Action, rp func(error) bool) Action {
+	return &retriableAction{a, rp}
+}
+
+// Run executes Action. If error is returned retry provider checks if Action
+// should be rerun
+func (ra *retriableAction) Run(ctx context.Context, c cloud.Cloud) (EventList, error) {
+	for {
```
bowei, line 37: The infinite loop concerns me. We should probably mandate that the caller supplies a maximum number of retries.

pkg/cloud/rgraph/exec/retrieable_action.go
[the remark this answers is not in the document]
bowei: There are standard backoff retry packages. I don't think it makes sense to push the rate limiting of retries down to the overall rate limit. What that will do is that if a operation is fast and retries right away, it will exhaust the rate limiter below -- potentially starving other operations.

The delay from the retry should be exposed here to avoid hiding that from the user.

pkg/cloud/rgraph/exec/retrieable_action.go
kl52752: done

pkg/cloud/rgraph/exec/retrieable_action.go
kl52752: added comment

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,54 @@
...
+
+// retriableAction is a decorator to the action adding retry mechanism
+type retriableAction struct {
+	Action
+	isRetriable func(error) bool
+}
+
+// NewRetriableAction decorates Action with retry function. On error the action
+// will be retried when isRetriable(err) == true.
+func NewRetriableAction(a Action, isRetriable func(error) bool) Action {
```
bowei, line 31: isRetriable is somewhat awkward => canRetry
Add duration return

pkg/cloud/rgraph/exec/retrieable_action.go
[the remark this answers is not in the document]
mag-kol: I agree with that. This is inline with what I wrote and I think Katarzyna is already working on fixing that.

kl52752:
> Discussed offline -- it seems like we should make the retry function return a duration so this can be rate limited.
> 
> As part of waiting for the duration, the code should also use a `switch` to watch for context cancellation.

this is done, PTAL

mag-kol (approved):

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		if backOffTime := ra.canRetry(err); backOffTime > 0 {
+			ticker := time.NewTicker(backOffTime)
+			select {
+			case <-ticker.C:
+				continue
+			case <-ctx.Done():
+				return nil, fmt.Errorf("context canceled")
+			}
+		}
```
bowei, line 55: you need to call ticker.Stop() otherwise you will have a leak

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+)
+
+// retriableAction is a decorator to the action adding retry mechanism
```
bowei, line 25: minor: this is not a decorate -- it is the action...

pkg/cloud/rgraph/exec/retrieable_action.go
```diff
@@ -0,0 +1,65 @@
...
+type retriableAction struct {
+	Action
+	canRetry func(error) time.Duration
+}
+
+// NewRetriableAction decorates Action with retry function. On error the action
+// will be retried when canRetry(err) returns duration greater than 0.
```
bowei, line 32: We should return (bool, time.Duration). Otherwise if you want to retry immediately (which could be applicable), the user will need to return epsilon time, which doesn't make sense.

kl52752: done

pkg/cloud/rgraph/exec/retrieable_action.go
kl52752: done

PR #206, opened 2024-05-21
Remove latency measurements from TickerRateLimiter tests

AwesomePatrol:
The first call returns without waiting for a full tick so it should take a minimum 490ms.

image

This would still be unreliable, so simplified the test to not rely on time.

### DIFF
--- a/pkg/cloud/ratelimit_test.go
+++ b/pkg/cloud/ratelimit_test.go
@@ -98,8 +98,10 @@ func TestTickerRateLimiter(t *testing.T) {
 	if elapsed > time.Second {
 		t.Errorf("TickerRateLimiter.Accept took too long: %v, want <1s", elapsed)
 	}
-	if elapsed < 500*time.Millisecond {
-		t.Errorf("TickerRateLimiter.Accept took too short: %v, want >500ms", elapsed)
+	// The first Accept call doesn't wait for the whole duration between
+	// ticks (it is possible it will return immediately).
+	if elapsed < 490*time.Millisecond {
+		t.Errorf("TickerRateLimiter.Accept took too short: %v, want >490ms", elapsed)
 	}
 
 	// Use context that has been cancelled and expect a context error returned.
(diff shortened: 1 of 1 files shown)


pkg/cloud/ratelimit_test.go
```diff
@@ -98,8 +98,9 @@ func TestTickerRateLimiter(t *testing.T) {
 	if elapsed > time.Second {
 		t.Errorf("TickerRateLimiter.Accept took too long: %v, want <1s", elapsed)
 	}
-	if elapsed < 500*time.Millisecond {
-		t.Errorf("TickerRateLimiter.Accept took too short: %v, want >500ms", elapsed)
+	// First request returns immediately.
```
kl52752, line 101: First request will be called immediately without delay not returned. Can you rephrase this?

kl52752:
/lgtm
/approve

Thanks for quick fix

pkg/cloud/ratelimit_test.go
```diff
@@ -87,27 +87,17 @@ func TestTickerRateLimiter(t *testing.T) {
...
-			t.Errorf("TickerRateLimiter.Accept = %v, want nil", err)
-		}
-	}
-	elapsed := time.Since(start)
-	if elapsed > time.Second {
-		t.Errorf("TickerRateLimiter.Accept took too long: %v, want <1s", elapsed)
-	}
-	if elapsed < 500*time.Millisecond {
-		t.Errorf("TickerRateLimiter.Accept took too short: %v, want >500ms", elapsed)
+	err := trl.Accept(context.Background(), nil)
```
kl52752, line 90: maybe you can reverse the logic, instead checking elapsed time just spin the go routine with trl.Accept for 500ms and check how many request were accepted?

kl52752: you mean, you removed time sensitive check not test right? 
sounds good

PR #207, opened 2024-05-29
Add region neg support

asiawildboar:
Add the omitted region network endpoint groups.


bowei:
Please use git rebase -i to clean up the commits:

You should have two commits:
 
* pkg/cloud/meta/compute_services.go
* generated files

The commit messages should look like this:

```
Add regional NEG to generated resource.
```

```
Regional NEG generated code.
```

PR #208, opened 2024-05-30
Add region neg support

asiawildboar:
Add the omitted region network endpoint groups.


bowei:
/ok-to-test
Let's retest and get your change in. I am going to fix that flaky test.

PR #209, opened 2024-05-31
Fixes the serial executor test

bowei:
* Make testAction respond to the context cancellation as documented in the docstring for serialExecutor.
* Pass context to the testAction so it can be used to cancel the blocking action.
* Adjust timers so the test is not as sensitive to timing issues.


bowei:
@kl52752 -- comments?

PR #214, opened 2024-07-12
Add run hash to bs resource avoid cleanup issues.

akwi-github:
Backend Services in tests were created without a unique prefix which caused issues with cleanup.

### DIFF
--- a/e2e/tcproute_test.go
+++ b/e2e/tcproute_test.go
@@ -288,28 +288,18 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
 	if err != nil {
 		t.Fatalf("buildNEG(_, neg-test, %s) = (_, %v), want (_, nil)", zone, err)
 	}
-	t.Cleanup(func() {
-		err := theCloud.NetworkEndpointGroups().Delete(ctx, negID.Key)
-		t.Logf("theCloud.NetworkEndpointGroups().Delete(ctx, %s): %v", negID.Key, err)
-	})
 
 	hcID, err := buildHealthCheck(graphBuilder, "hc-test", 15)
 	if err != nil {
 		t.Fatalf("buildHealthCheck(_, hc-test, 15) = (_, %v), want (_, nil)", err)
 	}
-	t.Cleanup(func() {
-		err := theCloud.HealthChecks().Delete(ctx, hcID.Key)
-		t.Logf("theCloud.HealthChecks().Delete(ctx, %s): %v", hcID.Key, err)
-	})
+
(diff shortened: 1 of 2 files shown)


e2e/tcproute_test.go
```diff
@@ -588,6 +588,7 @@ func TestMeshWithMultipleTCPRoutes(t *testing.T) {
 				t.Logf("delete TCProute: %v", err)
 			}
 		}
+		//bsss, err := theCloud.BackendServices().List(ctx, filter.None)
```
AwesomePatrol, line 591: this line looks useful

e2e/tcproute_test.go
```diff
@@ -346,6 +342,10 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, negID)},
 	}
 	processGraphAndExpectActions(t, graphBuilder, expectedActions)
+	t.Cleanup(func() {
```
kl52752, line 345: all clean-ups should be moved just before rGraph executed, otherwise no resources will not be created and there is no point to gc them. 
`buildHealthCheck` and `buildBackednServices` those functions are creating rNodes and adding them to rGraph but resource creation is done by executor.

e2e/tcproute_test.go
akwi-github: Done.

e2e/tcproute_test.go
akwi-github: Done.

kl52752: sorry I meant after executor

e2e/tcproute_test.go
```diff
@@ -334,18 +331,22 @@ func TestRgraphTCPRouteAddBackends(t *testing.T) {
...
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, bsID)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, hcID)},
 		{Type: exec.ActionTypeCreate, Name: actionName(exec.ActionTypeCreate, negID)},
 	}
+
 	processGraphAndExpectActions(t, graphBuilder, expectedActions)
+	t.Cleanup(func() {
+		var allErros error
+		allErros = errors.Join(allErros, theCloud.TcpRoutes().Delete(ctx, tcprID.Key))
+		t.Logf("theCloud.TcpRoutes().Delete(_, %s): %v", tcprID.Key, err)
```
kl52752, line 346: why do you need all errors if you print err? 
You should delete hc and negs here also

akwi-github: Done.

PR #222, opened 2024-08-21
Fix compression mode

akwi-github:
We need to leave compression mode empty as Compression mode can only be specified with EXTERNAL or EXTERNAL_MANAGED load balancing scheme.  In order to do so we need to allow zero value on the field.


akwi-github:
\cc @kl52752 @AwesomePatrol @mag-kol

akwi-github:
\cc @kl52752

bowei:
I'm approving the unblock the PR, but I'm not entirely certain the current type traits are doing what we need to do. I would appreciate if after we build out an initial set of use cases that we re-evaluate whether or not it supports what we need as currently implemented.

PR #225, opened 2024-09-03
Add Fallback Cleanup

AwesomePatrol:
In some cases (like the recent ensure test) it failed to clean up resources after the test run. Fallback mechanism will remove all resources from the project that match test prefix and `runID`.

`List` calls to `Meshes`/`TcpRoutes` are fixed in https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/226

Also, fixed some minor issues in tests.

Bug: b/352484069


AwesomePatrol:
Please let me know if you would prefer to gate this behavior behind a flag so it is used in CI only.

/cc @akwi-github 
/assign @kl52752

PR #226, opened 2024-09-03
Fix calls to NetworkServices List

AwesomePatrol:
It takes `parent` (which contains `projectID` and `location`) argument, but `projectID` was provided instead.

Bug: b/364244347

### DIFF
--- a/pkg/cloud/gen/main.go
+++ b/pkg/cloud/gen/main.go
@@ -837,7 +837,12 @@ func (g *{{.GCPWrapType}}) List(ctx context.Context, zone string, fl *filter.F, 
 
 {{- if .KeyIsGlobal}}
 	klog.V(5).Infof("{{.GCPWrapType}}.List(%v, %v): projectID = %v, ck = %+v", ctx, fl, projectID, ck)
+{{- if .IsNetworkServices }}
+	parent := fmt.Sprintf("projects/%s/locations/global", projectID)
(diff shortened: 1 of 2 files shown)


pkg/cloud/gen/main.go
```diff
@@ -837,7 +837,12 @@ func (g *{{.GCPWrapType}}) List(ctx context.Context, zone string, fl *filter.F,
 
 {{- if .KeyIsGlobal}}
 	klog.V(5).Infof("{{.GCPWrapType}}.List(%v, %v): projectID = %v, ck = %+v", ctx, fl, projectID, ck)
+{{- if not .IsNetworkServices }}
 	call := g.s.{{.GroupVersionTitle}}.{{.Service}}.List(projectID)
+{{- else}}
+	parent := fmt.Sprintf("projects/%s/locations/global", projectID)
```
bowei, line 843: Do we need to handle regional vs global?

AwesomePatrol: In the future, yes. I've seen implementation of Regional APIs just today - I need to asses how much work that may  need.
Currently I followed the existing behavior for `NetworkServices.Insert`: https://github.com/GoogleCloudPlatform/k8s-cloud-provider/blob/v1.33.0/pkg/cloud/gen/main.go#L918-L919

bowei: can we flip the conditional to avoid `not`

```
{{- if .IsNetworkServices}}

{{else}

PR #227, opened 2024-09-10
Change BackendService beta API tests

kl52752:
Fields present in beta.BackendService API might be promoted to v1 any time and it is hard to follow that changes. Having a test that checks that fields are present only in beta API and checkst that conversion between Ga and Beta version fails are unstable and hard to maintain. Removing this logic from tests.


bowei:
The whole point of the testing that conversion works is to make sure someone is paying attention and updates the code when the GA APIs are introduced. 

What is the mitigation on how are we going to track a change if we remove the test?

kl52752:
The problem with those test is that it is hard to maintain it across different version of API. The best way to test it is t really on e2e tests

PR #228, opened 2024-09-10
Add resource key to observer CallContext.

briantkennedy:
This allows the observer to know the location for the given operation.

### DIFF
--- a/pkg/cloud/context.go
+++ b/pkg/cloud/context.go
@@ -43,4 +43,6 @@ type CallContextKey struct {
 	Version meta.Version
 	// Service is the service being invoked (e.g. "Firewalls", "BackendServices")
 	Service string
+	// Resource is the key of the resource being operated on.
+	Resource *meta.Key
 }
--- a/pkg/cloud/gen/main.go
+++ b/pkg/cloud/gen/main.go
@@ -771,6 +771,7 @@ func (g *{{.GCPWrapType}}) Get(ctx context.Context, key *meta.Key, options... Op
 		Operation: "Get",
 		Version: meta.Version("{{.Version}}"),
 		Service: "{{.Service}}",
+		Resource: key,
 	}
 
 	klog.V(5).Infof("{{.GCPWrapType}}.Get(%v, %v): projectID = %v, ck = %+v", ctx, key, projectID, ck)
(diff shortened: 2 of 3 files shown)


pkg/cloud/context.go
```diff
@@ -43,4 +43,6 @@ type CallContextKey struct {
 	Version meta.Version
 	// Service is the service being invoked (e.g. "Firewalls", "BackendServices")
 	Service string
+	// Key is the key of the resource being operated on.
+	Key *meta.Key
```
kl52752, line 47: Can you rename it to a ResourceKey? CallContextKey.Key is not descriptive enough

bowei: +1, probably call it `Resource` to avoid stuttering.

briantkennedy: So...to begin with `Key` is already a strange name given that it represents a location for a resource, and not the resource itself.  I named it `Key` here for consistency since it's probably better to have a bad but consistent name than to use an arbitrary name because it clashes with names in the current package.  If we have a problem with `Key`, I think we need to alias / rename that type in the meta package to something more descriptive like `Location` or `Scope`.

briantkennedy: renamed to `Resource`

pkg/cloud/gen/main.go
```diff
@@ -810,16 +811,19 @@ func (g *{{.GCPWrapType}}) Get(ctx context.Context, key *meta.Key, options... Op
 func (g *{{.GCPWrapType}}) List(ctx context.Context, fl *filter.F, options... Option) ([]*{{.FQObjectType}}, error) {
         opts := mergeOptions(options)
 	klog.V(5).Infof("{{.GCPWrapType}}.List(%v, %v, %v) called", ctx, fl, opts)
+	key := &meta.Key{}
```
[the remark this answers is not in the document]
briantkennedy: It's already called `key` by convention in the method signature for `Get`,`Delete`, `Update`, and `Insert`.  Do we plan to change that as well?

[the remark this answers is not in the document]
kl52752: good point, no let's leave this

briantkennedy:
@bowei this needs your approval

briantkennedy:
@bowei

PR #236, opened 2025-01-10
Pr dynamic registration

bowei:


pkg/cloud/rgraph/rnode/all/new.go
```diff
@@ -18,40 +18,37 @@ package all
...
 )
 
+type newBuilderFunc func(id *cloud.ResourceID) rnode.Builder
+
+var (
+	builderFuncsLock sync.Mutex
+	builderFuncs     = map[string]newBuilderFunc{}
+)
+
+// RegisterBuilder constructor. shortName is the plural lowerCamelCase resource
```
kl52752, line 34: RegisterBuilder is not a constructor. The global variables are already created and this function is adding ("registering") the values so they can be used by rGraph. Can you change a bit the description?

pkg/cloud/rgraph/rnode/all/new.go
```diff
@@ -18,40 +18,37 @@ package all
...
-	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode/networkendpointgroup"
-	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode/targethttpproxy"
-	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode/tcproute"
-	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode/urlmap"
 )
 
+type newBuilderFunc func(id *cloud.ResourceID) rnode.Builder
+
+var (
+	builderFuncsLock sync.Mutex
```
kl52752, line 30: I think it will be useful ti use RWLock. Write will happened only in init. Then in rGraph we are reading the values from the map so RLock will be more efficient.

pkg/cloud/rgraph/rnode/backendservice/builder.go
```diff
@@ -23,11 +23,14 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/api"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/rgraph/rnode/all"
 	alpha "google.golang.org/api/compute/v0.alpha"
 	beta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 )
 
+func init() { all.RegisterBuilder("backendServices", NewBuilder) }
```
kl52752, line 32: please extract "backendServices" to a const so we can use the value from the package and make sure that no typo will be introduced in  `NewBuilderByID`. Please apply this comment to all cases.

pkg/cloud/rgraph/rnode/tcproute/builder.go
```diff
@@ -31,6 +32,8 @@ const (
 	resourceName = "TcpRoute"
 )
 
+func init() { all.RegisterBuilder("tcpRoute", NewBuilder) }
```
kl52752, line 35: should be plural    tcpRoute -> tcpRoutes

pkg/cloud/rgraph/rnode/all/new.go
bowei: The sentence is supposed to be "Register Builder constructor." but this is not allowed by go.. I will clarify by expanding the description.

PR #237, opened 2025-01-29
Add option for returning partial success when using AggregatedList.

briantkennedy:


pkg/cloud/option.go
```diff
@@ -21,6 +22,14 @@ func mergeOptions(options []Option) allOptions {
 	return ret
 }
 
+// ReturnPartialSuccess only affects AggregatedList calls, and will have the call return success
+// when fanout calls fail.
```
bowei, line 26: ...fanout to multiple locations...

